PR #2, opened 2016-01-16
Initial, work-in-progress code

bcronin:
Hi all -

Here's what I'll label the "work in progress" JavaScript code. It's far enough along that I've confirmed it works both as a no-op API as well as with (at least the basics of) our own backend. I am still actively working on this and will log some of the code-incomplete aspects of it as Github issues.

A couple high-level notes: 

* This is intended to be a single package that builds both a Node and browser distributable
* For both, this is intended to build both a debug version (with some debugging checks to help with OpenTracing API conformance) and a production version (minimal overhead, no checks, less debugging info)

To be a little more explicit about what's still work-in-progress for anyone looking at the code:

* Docs, the README, code comments, etc.: mostly placeholder at this point
* The build process is a little manual at the moment (i.e. the webpack config is undoubtedly sub-elegant)
* Some general confusion on how to implement the text and binary codecs without making platform/protocol/implementation assumptions
* Examples: I need to write some both for API use and for a backend implementor
* Unit tests, especially for the browser
* A unit-test like validator that can test API conformance for an arbitrary backend
* Benchmarking
* A few other things here and there, which I'll log in Github
* Any level of nitpicky self-review since I'm still actively working on this :)

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -1,0 +1,72 @@
+# Developer Notes
+
+This is a work-in-progress implementation. Feedback and contributions are welcome.
+
+This particular document is intended to summarize some development decisions made in the library to make moving the library forward easier and avoid revisiting past issues. The code history and Github issue history are helpful sources for more of this kind of information.
+
+### Strictness
+
+The library has a debug mode that enforces as many aspects of the public API as it reasonably can before passing data onto the implementation object. Often this is simply means validating the number and type of arguments. This is intended to discourage accidental (as well as intentional!) deviations in individual implementations from the OpenTracing specification. Particularly in debug mode, the library should err on the side of being overly restrictive (at least during initial development) as public APIs are almost always easier to make more flexible than more restrictive over time.
+
+### Errors and Exceptions
+
+The API layer throws Error exceptions *only* in debug builds.
+
+In spirit of keeping the instrumentation from ever interfering with production application logic, the API layer does not handle error conditions in production builds. OpenTracing implementations may choose to handle them, but the implementations are encouraged to adhere to the principle that instrumentation code should not adversely affect the primary application code.
+
+### No-op implementation
+
+A design choice has been made that, for the purposes of integration simplicity and thus ease-of adoption, the `opentracing` JavaScript package is both the common API as well as a no-op implementation. This means that including the package should not require explicit configuration of a no-op implementation backend in order to run without instrumentation.
+
+### API use of the Bridge Pattern ("pImpl idiom")
+
+The API layer uses a bridge pattern to pass work to the specific tracing implementation. This is done instead of providing direct access to implementation-specific objects.
+
+This is done primarily to encourage and, when possible enforce, standardization of the instrumentation code and conformance to the OpenTracing API. In JavaScript, which lacks a formal language-level concept of an "interface", the bridge pattern can emulate some of the advantages of more formal separation between interface and implementation.
+
+For example, it is almost inevitable that there will be use cases for particular implementations that cannot be handled directly by the OpenTracing API (both valid uses and misguided ones, no doubt!). Encapsulating the implementation object as separate from the standard API, at the very least, encourages the deviations from the standard to be made more explicit.
+
+For example, in the construed scenario below the `imp()` call below makes it self-documenting and obvious that `flushToStream()` is a non-standard, implementation-dependent method:
+
+```
+Tracer.startTrace("big_operation");
+Tracer.imp().flushToStream(stream);
+```
+
+As a practical matter, such explicitness can be helpful for those new to tracing code (it self-documents what is standard and what is implementation-specific) as well as to reviewers of such code in highlighting deviations from the standard that may creep into a codebase.
+
+## Open Questions / Issues / TODOS
+
+*Note: the unresolved issues should be migrated to Github once the `opentracing-javascript` repository is created.*
+
+**Don't the argument checks created overhead?**
+
+Yes. Since webpack is being used, the code should be modified so these are compiled out of the production version.  A Github issue should be opened on to properly exclude these from production builds of the package.
+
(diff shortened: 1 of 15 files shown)


DEV.md
```diff
@@ -0,0 +1,72 @@
+# Developer Notes
```
bensigelman: per https://github.com/opentracing/opentracing.github.io/issues/10, point to `opentracing.io` for general background on the project, semantics, etc.

README.md
```diff
@@ -1,2 +1,81 @@
...
+```js
+var Tracer = require('opentracing');
+var TracingBackend = require('tracing-implementation-of-your-choice');
+
+Tracer.initGlobalTracer(TracingBackend.create());
+```
+
+## Concepts
+
+The main OpenTracing project (http://opentracing.github.io) provides the most up to date documentation on concepts, semantics, and best practices around effectively using OpenTracing.
```
bensigelman: oh, sorry -- maybe disregard my other comment (or add a link there, too).

src/tracer.js
```diff
@@ -0,0 +1,181 @@
...
+    imp() {
+        return this._imp;
+    }
+
+    /**
+     * TODO: Unclear what the OpenTracing specification wants here.  This
+     * implementation is a best-guess approximation that's likely not right.
+     * The spec refers to text and binary representations, but what does
+     * "binary" mean in JavaScript without making assumptions about protocols,
+     * transports, and platforms?
```
bcronin: As the code comment here indicates, I was scratching my head a bit as to what the intention of the TraceContext encode / decode *specifically* to text and binary formats was for. Put differently, I wholeheartedly understand and support the need for the ability to serialize a TraceContext, but didn't quite understand the binary case (in particular) well enough to know how to implement this API in a general fashion in JS.

That's likely best discussed as part of the spec documentation rather than here, but wanted to note it here as this code undoubtedly needs to be revised once I understand the intention a bit better!

bensigelman: Agreed that the spec should be clearer... I'll make an appropriate change in that pending spec PR.

"binary" means an arbitrary-length byte array (as opposed to a utf8 string). "text" means a `string->string` map in whatever data structure is conventional. See also: https://github.com/opentracing/opentracing-java/pull/4/files#diff-1f61a455f742fcd354fedb93c07d28e8R24

src/span.js
```diff
@@ -0,0 +1,144 @@
+'use strict';
+
+import * as Constants from './constants';
```
yurishkuro: I'm not a JS expert, but most of our internal Node code I've seen uses `require`, I've never seen imports like this. Is it really universal syntax?

cc @oibe

src/span.js
```diff
@@ -0,0 +1,144 @@
...
+    imp() {
+        return this._imp;
+    }
+
+    // ---------------------------------------------------------------------- //
+    //
+    // ---------------------------------------------------------------------- //
+
+    traceContext() {
+        if (Constants.API_CONFORMANCE_CHECKS) {
```
yurishkuro: can we do API conformance check in a decorator class instead of the core class?

src/span.js
bcronin: This is [ES6 syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import). 

The JS is being compiled with [Babel](https://babeljs.io/) so there's less concern about what particular browser or Node version is being used by the consumer of the code.

src/tracer.js
bcronin: The problem is that, as far as I know, there's no standard "byte array" in JavaScript as a language. Yes, really :)

* All of JS supports an `Array` of `number` objects - but this is very inefficient
* Node.js supports `Buffer` (largely to compensate for this)
* The browser does *not* support Buffer
* Some (most?) browsers support [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array), but I don't know how common this is or how well it would adapt to the use case that this binary format is being required for

https://github.com/opentracing/opentracing-javascript/issues/6 might be a good place to discuss this issue.

src/span.js
bcronin: The intention is that these are zero-overhead in production. `API_CONFORMANCE_CHECKS` will be `false` in production -- and since there is a compile step, this will be dead code (not an `if (false) {`) check.  For simplicity sake, I like the idea that the conformance checks are always part of the core debug build.

(And for the record, this isn't the reality yet -- still a tiny bit of work to do in the `webpack.config.js` to do that.)

webpack.config.js
```diff
@@ -0,0 +1,109 @@
...
+
+switch (CONFIG) {
+    case "debug":
+        defines.DEBUG = true;
+        bundleSuffix = "-debug";
+        devtool = "source-map";
+        break;
+
+    case "prod":
+        plugins.push(new webpack.optimize.UglifyJsPlugin({
```
yurishkuro: do we want this for Node?

bcronin: Yes, I believe we do. The short answer in the general case is that minified code (still) tends to run faster on Node than unminified code (an example is React.js), to the best of my understanding at least. In the specific case of this code, It's relying on the dead code elimination of this compilation step to ensure those argument checks truly are zero ovehead (i.e. non-existent!). The newest versions of Node/V8 may be smart enough to do that natively, but `const` actually hasn't been around that long in JS so I would be hesitant to rely on true constant-propogation and code path elimination in older versions of V8.

bcronin:
Are there are blockers to my merging this? Again, it's definitely WIP, so I'm glad to revise anything with a subsequent pull request.

@yurishkuro, let me know if you'd like me to wait for @oibe to chime in before proceeding!

yurishkuro:
lgtm, :ship:

PR #9, opened 2016-01-23
Make initNewTracer() mirror initGlobalTracer()

bcronin:
Changes the signature of `initNewTracer()` to take an tracer implementation object (rather than implicitly create a new one via the global tracer.  I believe this qualifies as a small, logical change as it makes two similar methods have similar signatures.

Note: the OpenTracing spec at this point specifies the `Tracer` interface but not how tracers are created / acquired; with that in mind, I believe this change fits with the standard.

(Also for what it's worth, the LightStep tracer allows multiple Tracer instances to be created, so this change was needed to support that functionality.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opentracing",
-  "version": "0.9.0",
+  "version": "0.9.1",
   "main": "dist/opentracing-node-debug.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


bcronin:
Given the uncontentious nature of this change, I'm going to go ahead and merge it (to help me move forward with validating OpenTracing against the LightStep JS implementation).  

If I'm wrong about it being uncontentious, I'll gladly back it right out again! :)

bensigelman:
Anything that brings us closer to being a "pure" OT shop internally is auto-merge in my book!

Issue #7, opened 2016-01-19
Production build does not (yet) remove argument checks


bcronin:
The webpack / babel configuration for production should set the `API_CONFORMANCE_CHECKS` constant to false and ensure the resulting dead code is remove (for performance reasons).

This is not yet implemented.

PR #10, opened 2016-01-25
Run API conformance checks in debug only

bcronin:
Ensure the `API_CONFORMANCE_CHECKS` code blocks (i.e. check number and type of arguments on each call) exist only in debug.

Specifically, modifies the `webpack.config.js` configuration to set `API_CONFORMANCE_CHECKS` as a constant that will be removed by dead-code elimination step of the UglifyJS plugin. Note: I manually verified in the generated production and debug code that the checks will be done only in debug builds.

Resolves https://github.com/opentracing/opentracing-javascript/issues/7.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,6 +1,0 @@
-'use strict';
-
-// TODO: make this into a webpack build option so that production builds do
-// not necessarily need to include the overhead of the argument checks.
-// For the initial iterations, default to this being enabled.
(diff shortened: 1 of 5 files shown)


bcronin:
Moving forward with the merge here as I believe this pull request is pretty uncontentious.  

Glad to revert / revise if there's an issue. Just let me know!

PR #11, opened 2016-02-19
Update to latest spec

bcronin:
This is an incremental update to the `opentracing-javascript` code to get more in line with the current `opentracing-python` / `opentracing-go` public APIs. I'm qualifying it as "incremental" in that the docs are still out of date in various places and I'm still working on migrating the LightStep code to validate this with a real backend implementation; however, this gets the JS API layer a lot closer to the status quo.


bcronin:
I'm going to proceed with the merge as I believe this is uncontentious (i.e. it's updating the code to match the existing Go / python APIs and not introducing anything new).  Glad create a follow-up if there are any issues :)

PR #12, opened 2016-02-24
Automate webpack to run as part of publish step

bcronin:
Automates the JavaScript compilation step (i.e. running `webpack`) when publishing the OpenTracing NPM package.

The two changes are:

1. Include an empty `.npmignore` in the `dist` directory so the published package *does not* ignore the files in the `dist` directory. The default behavior of NPM is to use the `.gitignore` file if there is no `.npmignore` file, but in this case, the files in the published package and the files in the source are *not* supposed to be the same.

2. Add a `prepublish` "script" step that runs the `webpack` step to do the build. The `prepublish` script is automatically invoked on publish so that building the source is no longer a manual step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: node_js
+node_js:
+  - 0.12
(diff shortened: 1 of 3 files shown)


bcronin:
Assuming from the lack of commentary that no one has any issues with this. Merging this update.

PR #14, opened 2016-03-02
Sync to recent OpenTracing spec updates

bcronin:
Gets the JavaScript code in sync with some of the more recent OpenTracing spec updates:

- Renames the `set/getTraceAttribute` APIs to `set/getBaggageItem`
- Adds the binary and split text carriers and `inject`/`join` APIs to `Tracer`
- Adds `setOperationName` to `Span` (not new to the spec, but was not in the JS code before)
- Return the `this` object by default on most `Span` methods without an explicit return value in case users want to chain calls (and more specifically, since that's the pattern Python seems to be doing and keeping the platform APIs similar seems like a good thing)

  ​

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This library is a JavaScript implementation of Open Tracing API. It is intended 
 
 ## Required Reading
 
-In order to fully understand this platform API, one must must first be familiar with the [OpenTracing project](http://opentracing.io) and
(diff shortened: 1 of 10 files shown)


bcronin:
Since there shouldn't be anything contentious in here, moving forward with the merge.

PR #15, opened 2016-03-04
Tracer method

bcronin:
Two changes:

* Adds the `Span.tracer()` method to the API
* Removes the dead code based on the old Injector/Extractor interfaces


bcronin:
Going to check this in directly as it addresses a clear spec omission from the last PR and is pretty straightforward.

PR #16, opened 2016-03-05
API conformance

bcronin:
Fixes a few small divergences from the common API:

* Removes `startChildSpan` from `Span`
* Fix two references to `TraceAttributes` that should have been `Baggage`
* Allow `startSpan(operationName, { parent : parent})` as a call signature


bcronin:
(Going to merge this since it's essentially just a few corrections to the last pull request)

PR #19, opened 2016-03-18
Remove double Tracer implementation initialization

bcronin:
## Summary

Removes an accidental "double initialization" when setting the global tracer. Rather than using the passed-in object directly, it was cloning the passed in object via a call to `newTracer`.  This "worked" so wasn't noticed before, but is obviously inefficient!

Also, this changes the webpack config to use "umd" format (rather than "var" format) for the browser for broader compatibility.  The "var" format makes assumptions about accessibility of global variables that may not be true in different JS module systems.

### DIFF
--- a/src/singleton.js
+++ b/src/singleton.js
@@ -20,9 +20,11 @@ export default class Singleton extends Tracer {
      * Set the global Tracer's underlying implementation.
      *
      * The behavior is undefined if this function is called more than once.
+     *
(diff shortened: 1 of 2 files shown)


bcronin:
As this pull request addresses a rather obvious (in retrospect) defect, I'm going to proceed directly with the merge.  Glad to address any concerns with a quick follow-up if I'm making any incorrect assumptions about the obviousness of the fix here!

PR #20, opened 2016-03-23
Move to the simplified carriers

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/73

@bcronin


src/tracer.js
```diff
@@ -91,6 +89,20 @@ export default class Tracer {
...
+     * Consider this pseudocode example:
+     *
+     *     var clientSpan = ...;
+     *     ...
+     *     // Inject clientSpan into a text carrier.
+     *     var textCarrier = {};
+     *     Tracer.inject(clientSpan, Tracer.FORMAT_TEXT_MAP, textCarrier);
+     *     // Incorporate the textCarrier into the outbound HTTP request header
+     *     // map.
+     *     outboundHTTPReq.headers.extend(textCarrier);
```
bensigelman: Open to suggestions for the pseudocode here (variable names, etc)... but I do like the idea of providing a simple example. See also the join() example.

bcronin:
It looks like there are still other references to `FORMAT_SPLIT_TEXT` in the code.  E.g. https://github.com/opentracing/opentracing-javascript/pull/20/files#diff-a1434a31ad8190344d670eddd98c80ffR124

bensigelman:
yikes, sorry... I added a test as penance.

bcronin:
Cool.  Looks good to me.  Thanks for getting this up to date with the OpenTracing spec!

bcronin:
@bensigelman , I might be wrong, but I'm not sure the signature is going to work for JavaScript:

```js
inject(span, format, carrier)
```

...where `carrier` is an `ArrayBuffer`.

[ArrayBuffers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) are fixed-length at construction time. This signature would require the caller to know the size of the buffer the implementation needs -- which, unless there's something I'm missing, is not a reasonable expectation :)

This could easily be fixed by having `inject` *return* a carrier rather than have one passed in.  I'm not sure if this is a "valid" platform-specific variation from the semantic spec or not.

Another option is that the binary carrier is defined as a plain `Object` with a single `buffer` field that the implementation sets.

I'm not aware of a native resizable binary buffer type that's broadly available in browsers that could be used in place of the fixed-length Array types to keep the inject signature as-is.

yurishkuro:
It could be just an object with a slot for the array buffer to be set by the injector. I don't think returning the carrier is a good approach, how would that work if the carrier is some sort of RPC request object?

bcronin:
@yurishkuro , moving the conversation to #22.  (Seems a bit better to have it as part of an issue rather than in this closed pull request!)

PR #21, opened 2016-03-25
Improve compatibility

bcronin:
## Summary

* Removes Babel ES2015 usage of `Symbol`: this (a) is not needed in the OpenTracing layer for anything and (b) causes problems in some versions of IE (thanks @luk3thomas for pointing this out!)
* Adds code to further enforce the tracer singleton really is a singleton. Depending on the client code set up, this now captures some cases where multiple singletons could end up being created (e.g. OpenTracing is both included explicitly as a `<script>` on a page and is implicitly included via some build process like browserify in the binding library)
* Adds a Makefile for convenience (e.g. build the code if necessary before running the tests)


bcronin:
Merging with assumption that there's nothing contentious here. Glad to follow-up if there's that assumption was incorrect!

Issue #22, opened 2016-03-25
inject() method signature does not work well with ArrayBuffer as a carrier


bcronin:
(Copy-and-paste of a pull request comment below)

FYI, @bensigelman  making this into a separate issue since the PR was already merged.

---- 

@bensigelman , I might be wrong, but I'm not sure the signature is going to work for JavaScript:

```js
inject(span, format, carrier)
```

...where `carrier` is an `ArrayBuffer`.

[ArrayBuffers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) are fixed-length at construction time. This signature would require the caller to know the size of the buffer the implementation needs -- which, unless there's something I'm missing, is not a reasonable expectation :)

This could easily be fixed by having `inject` *return* a carrier rather than have one passed in.  I'm not sure if this is a "valid" platform-specific variation from the semantic spec or not.

Another option is that the binary carrier is defined as a plain `Object` with a single `buffer` field that the implementation sets.

I'm not aware of a native resizable binary buffer type that's broadly available in browsers that could be used in place of the fixed-length Array types to keep the inject signature as-is.

bcronin:
(Moving the conversation from the closed pull request to here)

From @yurishkuro :

> It could be just an object with a slot for the array buffer to be set by the injector. I don't think returning the carrier is a good approach, how would that work if the carrier is some sort of RPC request object?

bcronin:
@bensigelman also commented returning a carrier is not valid. So, I'm convinced the return value approach is not the right one :)

Perhaps the JavaScript-specific inject `FORMAT_BINARY` semantics as something like this:

--- 

* `carrier` needs to be a valid `Object`.  `inject` will set `carrier.buffer` to a `ArrayBuffer` containing the injected binary data.  Any valid `Object` can be used as long as the `buffer` field of the object can be set.

Optionally, it could also say (though I don't know if this additional bit is required; if it isn't I'd like to drop it to keep things simple):

* If the passed in `carrier` already has `carrier.buffer` field set to an `ArrayBuffer`, then `inject` will set the field to a new `ArrayBuffer` which contains the concatenation of the prior value and the newly injected binary data

Lastly, the `opentracing-javascript` library would also include an `opentracing.BinaryCarrier` class that's basically an empty object with a single `buffer` field - just for convenience for the simple case.

---

[...]

PR #23, opened 2016-03-25
Update FORMAT_BINARY carrier format

bcronin:
## Summary

Resolves https://github.com/opentracing/opentracing-javascript/issues/22 by updating the semantics of the `carrier` for `FORMAT_BINARY`.

In short, `carrier` was previously specified as being an `ArrayBuffer` - but in JavaScript those are fixed length buffers, which creates some usage complications.  The specification is now "loosened" to be any valid object with a `buffer` field - which in turn can any array-like type (e.g. ArrayBuffer, Array, Uint8Array, Float32Array, etc.)

### DIFF
--- a/src/binary_carrier.js
+++ b/src/binary_carrier.js
@@ -1,0 +1,13 @@
+'use strict';
+
+/**
+ * Convenience class to use as a binary carrier.
+ *
+ * Any valid Object with a field named `buffer` may be used as a binary carrier;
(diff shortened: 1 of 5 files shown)


bcronin:
@yurishkuro , since you're on the thread for #22, want to take a glance at this before I merge?

yurishkuro:
lgtm, modulo my limited knowledge of JS

PR #24, opened 2016-03-29
Handle the case of <script> and webpack

bcronin:
## Summary

Fixes a defect where the singleton was not properly shared when the library consumer was including OpenTracing, for example, both via a `<script>` tag and an implicit copy via webpack or browserify.

The code simply incorrect before as it never set the shared singleton value.

### DIFF
--- a/src/default_tracer.js
+++ b/src/default_tracer.js
@@ -6,12 +6,15 @@ import Singleton from './singleton';
 // tags vs included via webpack), it's possible for a single application to
 // have multiple copies of the OpenTracing code.  Try to ensure there's only one
 // singleton even in such scenarios.
-var singleton;
-if (typeof window !== 'undefined' && window.__opentracing_singleton) {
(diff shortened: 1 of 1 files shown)


bcronin:
As this is a fairly self-evident defect, I'm going ahead with the merge. As usual, I'm glad to address any concerns I might have missed here with follow-ups!

syrnick:
This causes a global leak:

Error: global leak detected: __opentracing_singleton (when running tests with mocha --check-leaks ...)

In node, the libraries should not use global namespace, module itself is a singleton, so it's a good place to stash the singleton. Sometimes it's tricky to ensure that the module is really a singleton due to the relative requires, but then it's up to the consumer to ensure that it's required consistently.

bcronin:
Thanks @syrnick . Moved this to issue https://github.com/opentracing/opentracing-javascript/issues/25.

Issue #25, opened 2016-03-29
Shared singleton causes a global leak


bcronin:
From @syrnick in #24 :

> This causes a global leak:
> 
> Error: global leak detected: __opentracing_singleton (when running tests with mocha --check-leaks ...)
> 
> In node, the libraries should not use global namespace, module itself is a singleton, so it's a good place to stash the singleton. Sometimes it's tricky to ensure that the module is really a singleton due to the relative requires, but then it's up to the consumer to ensure that it's required consistently.

bcronin:
(Preface: I don't claim to be an expert of how NPM module inclusions work! I am glad to be corrected here if I'm making wrong assumptions.)

Ensuring the OpenTracing instance is a single, shared object may be complicated. In the fully general case, I don't think the statement the "module itself is a singleton" is always true (at least in the *logical* sense of OpenTracing being one module). For more detail, see http://justjs.com/posts/singletons-in-node-js-modules-cannot-be-trusted-or-why-you-can-t-just-do-var-foo-require-baz-init 

OpenTracing, as a goal, hopes to be integrated into lots of packages. Being bit forward-looking, that means that it's quite possible that a single application might have two inclusions of OpenTracing and at different (compatible) versions.  These nonetheless may want to share a single underlying tracer implementation (i.e. a singleton).  

I think there are some considerations here that overlap with those in https://github.com/opentracing/opentracing-javascript/issues/8.

bcronin:
@synrick, what version of mocha are you using?

I certainly believe mocha is complaining about the use of window / global, but I'm (surprisingly) unable to reproduce the error with with either `--check-leaks` on the command line or `mocha.checkLeaks()` in the browser. I'm using v2.4.5. I'd certainly like to be able to reproduce the error before attempting to fix it.

bcronin:
@syrnick , sorry for the noise: it's something particular to my system (though I'm not sure what). I can reproduce the leak inside a clean Docker container.

bcronin:
@syrnick: I'm still looking for a way to reproduce the mocha error locally.  Is this some interaction with your build process or there is a way to recreate the error directly in this repo? I created a very basic unit test in #26 and it is passing on my system as well as the TravisCI build. 

I'm still of the mind that using a global is correct here (a cross-module singleton *really is* desired here) so I'd like to address this by avoiding the Mocha error not removing the single implementation instance.

syrnick:
Here's a few thoughts:

[...]

Issue #8, opened 2016-01-20
Ensure forward-compatibility scenario works correctly


bcronin:
Consider the following scenario in Node:

* Application A has a dependency on package B
* A has a dependency on `opentracing@1.1.0`
* B has a dependency on `opentracing@1.0.0`

All trace data recorded with in package B should be reported to (and be compatible) with the implementation set up in Application A.

This situation contains quite a few subtleties. For example, there will be two versions of opentracing within the global application and they need to ensure they share the same tracing implementation object (so data is all funneled to the same place).  Explicitly ensuring this scenario is accounted for would be good since it's (unsurprisingly) very difficult to correct forwards-compatibility issues after the fact.

bcronin:
Closed this a bit prematurely. I do think this is a scenario that needs to be considered. Implementing a single shared tracer instance across different modules may be not-that-obvious.  E.g. http://fredkschott.com/post/2013/12/node-js-cookbook---designing-singletons/

PR #26, opened 2016-03-30
Add test for mocha global leak

bcronin:
## Summary

* Adds the `--check-leaks` flag to the mocha test runner
* Adds a Make target to run the tests in Docker containers for various versions of Node

@syrnick , I'm adding this with regards to https://github.com/opentracing/opentracing-javascript/issues/25.  I'm trying to find a reproducible case of the mocha leak report. Could you run this branch / suggest a test to report the error?


bcronin:
### Update on this PR

* Removes the "global singleton" handling (see notes below)
* Adds a setInterface() method for the binding library (see notes below)
* Adds an *explicit* "NoOp" implementation binding for testing purposes
* Adds the `--check-leaks` flag to Mocha to be more strict
* Adds a unit test to capture the prior Mocha leak
* Adds some (optional) Docker-based tests to test against multiple Node versions

*Notes*

There was a lot of conversation on the usage of the global singleton. I do think there are some not-that-unrealistic scenarios where use of a global singleton could be beneficial, in very least by *detecting* multiple inclusions of the library rather than silently failing as the "global tracer" ends up being bound to only part of the dependency tree. However, I'll step off the soapbox and leave that to be resolved in #8 :)

This change also adds a `setInterface(tracer)` method on to the *binding* library. The requirement for this results from the desire to support the scenario where a OpenTracing consumer is *not* using a script bundler (like webpack or browserify) and includes both OpenTracing and the binding library as `<script>` tags. The binding library needs to be passed a handle to the OpenTracing object that consumer of the libraries is using. With the global singleton removed, this has be done explicitly via call to binding library. Otherwise if the binding library imports its own copy of OpenTracing, it's going to have a handle to a different instance; if it doesn't include its own copy, then it needs to rely on a global which isn't there anymore.

bensigelman:
@bcronin Makes sense... FWIW, this reminds me a bit of the [idiom for jquery plugin init](https://learn.jquery.com/plugins/basic-plugin-creation/#protecting-the-alias-and-adding-scope) (which was contending with a similar problem).

PR #28, opened 2016-04-05
Add conditionals to allow a null implementation object

bcronin:
## Summary

Allow `let tracer = OpenTracing.initNewTracer(null)` to be a valid way to get a no-op tracer instance.

* Fix a defect where the code assumed a non-null value for e tracer
* Add a unit test to ensure passing null creates a valid no-op tracer

Note: the majority of the diffs are cosmetic indentation changes to the unit tests.


bcronin:
As this is essentially just a "check for null", I'm going to proceed with the merge. Glad to discuss if there are any issues here!

PR #29, opened 2016-05-14
change implementation to receive `imp` instance

CatTail:


src/tracer.js
```diff
@@ -80,6 +80,9 @@ export default class Tracer {
             } else {
                 fields.operationName = nameOrFields;
             }
+            if (fields.parent) {
+                fields.parent = fields.parent._imp;
```
yurishkuro: I think this can be done in the implementation. I am not a fan of a function modifying the dictionary it receives as a parameter. For that matter, not happy with L81, but it's a different discussion (on gitter).

CatTail: Of course we can do this in implementation, but the implementation should always receive arguments it implement

yurishkuro: I could argue the opposite, that implementation should always handle the arguments of the _interface it implements_, and convert them to internal implementation as needed.

That's not a sticking point to me, but the fact that we're changing the dictionary passed as argument is. I'd prefer an API that does not have side effects, so the following would've made more sense to me:

```javascript
startSpan(operationName, fields) {
  if (API_CONFORMANCE_CHECKS) { ... }
  return new Span(this._imp.startSpan(operationName, fields));
}
```

CatTail: Internal implementation have no idea where interface store implementations unless dig into the source code. And interface may forbid outside world to access it's internal implementation!

Take a look at https://github.com/opentracing/opentracing-javascript/blob/master/src%2Ftracer.js#L138, at least we should be consistency, isn't it?

Another thought about modify arguments, there are simple way to fix this

```js
startSpan(operationName, fields) {
  if (API_CONFORMANCE_CHECKS) { ... }
  return new Span(this._imp.startSpan({
    operationName: operationName,
    parent: fields.parent._imp,
    // other fields
  }));
}
```

CatTail: @yurishkuro Maybe a real implementation can explain my idea. Take a look at this https://github.com/CatTail/basictracer-javascript/blob/fa6e42d4228fa922ddc9e0c96ff204b1c17153cc/src/__tests__/index-test.js#L48

src/tracer.js
```diff
@@ -70,17 +70,19 @@ export default class Tracer {
...
+                    nameOrFields = fields.operationName;
+                } else {
+                    fields = {};
                 }
-            } else {
-                fields.operationName = nameOrFields;
             }
-            spanImp = this._imp.startSpan(fields);
+            spanImp = this._imp.startSpan({
+                operationName: operationName,
```
bensigelman: you mean `operationName: nameOrFields`?

src/tracer.js
```diff
@@ -70,17 +70,19 @@ export default class Tracer {
...
+                parent: fields.parent && fields.parent._imp,
```
bensigelman: .js is not my most familiar language, so apologies if this is an obvious question: what is the `&& fields.parent._imp` for?

bensigelman:
@bcronin it would be great if you could take a look at this when you get a chance.

CatTail:
@bensigelman Just fix issues you metioned.

```js
fields.parent && fields.parent._imp
```

is equal to

```js
fields.parent ? fields.parent._imp : undefined
```

I just change the code style to make it more readable.

src/tracer.js
```diff
@@ -70,17 +70,19 @@ export default class Tracer {
...
+                } else {
+                    fields = {};
                 }
-            } else {
-                fields.operationName = nameOrFields;
             }
-            spanImp = this._imp.startSpan(fields);
+            spanImp = this._imp.startSpan({
+                operationName: nameOrFields,
+                parent: fields.parent ? fields.parent._imp : undefined,
```
bcronin: I understand the change to have the implementation's startSpan get a `imp` object rather than the interface object, however it's unclear to me that this is a "fix" rather than just a "change." Does the prior behavior somehow prevent some code from working? (I assume no, since the called code could always reference `parent._imp`.) Is this more consistent with other code in the package?

CatTail: 1. It's more consistent with other code in the package
2. Implementation expect receive `imp` rather then interface. I think "change" is more properly then "fix"

bcronin: It's not very important and I don't feel strongly about this, but I'd rather leave the code as it is if this PR is not addressing a particular issue or defect.

I think there's a *weak* argument it's better to pass the span interface as, within the code implementation can always reference the span's implementation but it cannot do the opposite (i.e. reference the span interface given the span implementation object). Also, the LightStep implementation, for example, [relies on the old behavior](https://github.com/lightstep/lightstep-tracer-javascript/blob/master/src/imp/span_imp.js#L173). The LightStep code would be easy to update if you feel strongly that about this change.

PR #33, opened 2016-07-08
A SpanContext RFC for Javascript

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/99

cc: @bcronin

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,9 @@
+.. :changelog:
+
+History
+-------
+
+0.10.0 (unreleased)
+-------------------
+
+- Move to SpanContext, Reference, and `s/join/extract/g`
--- a/src/reference.js
+++ b/src/reference.js
@@ -1,0 +1,38 @@
+'use strict';
+
+/**
+ * Reference pairs a reference type constant (e.g., REFERENCE_CHILD_OF or
+ * REFERENCE_FOLLOWS_FROM) with the SpanContext it points to.
+ *
+ * See the exported childOf() and followsFrom() functions at the package level.
+ */
+export default class Reference {
+
+    /**
+     * @return {string} The Reference type (e.g., REFERENCE_CHILD_OF or
+     *         REFERENCE_FOLLOWS_FROM).
+     */
+    type() {
+        return this._type;
+    }
+
+    /**
+     * @return {SpanContext} The SpanContext being referred to (e.g., the
+     *         parent in a REFERENCE_CHILD_OF Reference).
+     */
+    spanContext() {
+        return this._spanContext;
+    }
+
+    /**
+     * Initialize a new Reference instance.
+     *
+     * @param {type} string - the Reference type constant (e.g.,
+     *        REFERENCE_CHILD_OF or REFERENCE_FOLLOWS_FROM).
+     * @param {spanContext} SpanContext - the SpanContext being referred to.
+     */
+    constructor(type, spanContext) {
+        this._type = type;
+        this._spanContext = spanContext;
+    }
+}
--- a/src/reference.js
+++ b/src/reference.js
@@ -27,9 +27,9 @@ export default class Reference {
     /**
      * Initialize a new Reference instance.
      *
-     * @param {type} string - the Reference type constant (e.g.,
+     * @param {string} type - the Reference type constant (e.g.,
      *        REFERENCE_CHILD_OF or REFERENCE_FOLLOWS_FROM).
-     * @param {spanContext} SpanContext - the SpanContext being referred to.
+     * @param {SpanContext} spanContext - the SpanContext being referred to.
      */
     constructor(type, spanContext) {
         this._type = type;
--- a/src/reference.js
+++ b/src/reference.js
@@ -1,5 +1,8 @@
 'use strict';
 
+import Span from './span';
+import SpanContext from './span_context';
+
 /**
  * Reference pairs a reference type constant (e.g., REFERENCE_CHILD_OF or
  * REFERENCE_FOLLOWS_FROM) with the SpanContext it points to.
@@ -29,10 +32,20 @@ export default class Reference {
      *
      * @param {string} type - the Reference type constant (e.g.,
      *        REFERENCE_CHILD_OF or REFERENCE_FOLLOWS_FROM).
-     * @param {SpanContext} spanContext - the SpanContext being referred to.
+     * @param {SpanContext} spanContext - the SpanContext being referred to. As
+     *        a convenience, a Span instance may be passed in instead (in which
+     *        case its .context() is used here).
      */
     constructor(type, spanContext) {
+        if (API_CONFORMANCE_CHECKS) {
+            if (!(spanContext instanceof SpanContext || spanContext instanceof Span)) {
+                throw new Error('spanContext must be a Span or SpanContext instance');
+            }
+        }
         this._type = type;
-        this._spanContext = spanContext;
+        this._spanContext = (
+                spanContext instanceof Span ?
+                spanContext.context() :
+                spanContext);
     }
 }
--- a/src/singleton.js
+++ b/src/singleton.js
@@ -3,6 +3,7 @@
 import Tracer from './tracer';
 import * as Constants from './constants';
 import BinaryCarrier from './binary_carrier';
+import Reference from './reference';
 
 /**
  * The Singleton object is the default export of the package and extends the
@@ -44,6 +45,28 @@ export default class Singleton extends Tracer {
             tracerImp.setInterface(this);
         }
         return tracer;
+    }
+
+    /**
+     * Return a new REFERENCE_CHILD_OF reference.
+     *
+     * @param {SpanContext} spanContext - the parent SpanContext instance to
+     *        reference.
+     * @return a REFERENCE_CHILD_OF reference pointing to `spanContext`
+     */
+    childOf(spanContext) {
+        return new Reference(Constants.REFERENCE_CHILD_OF, spanContext);
+    }
+
+    /**
+     * Return a new REFERENCE_FOLLOWS_FROM reference.
+     *
+     * @param {SpanContext} spanContext - the parent SpanContext instance to
+     *        reference.
+     * @return a REFERENCE_FOLLOWS_FROM reference pointing to `spanContext`
+     */
+    followsFrom(spanContext) {
+        return new Reference(Constants.REFERENCE_FOLLOWS_FROM, spanContext);
     }
 
     // ---------------------------------------------------------------------- //
--- a/src/singleton.js
+++ b/src/singleton.js
@@ -92,6 +92,7 @@ export default class Singleton extends Tracer {
         for (let key in Constants) {        // eslint-disable-line no-restricted-syntax
             this[key] = Constants[key];
         }
+        this.Reference = Reference;
 
         // Carrier objects to be exposed at the package level
         this.BinaryCarrier = BinaryCarrier;
--- a/src/span.js
+++ b/src/span.js
@@ -31,7 +31,6 @@ export default class Span {
         if (this._imp) {
             spanContextImp = this._imp.context();
         }
-        // If there's no imp...
         return new SpanContext(spanContextImp);
     }
 
--- a/src/span.js
+++ b/src/span.js
@@ -1,9 +1,8 @@
 'use strict';
 
 import Tracer from './tracer';
+import SpanContext from './span_context';
 let defaultTracer = require('./default_tracer');
-
-const kKeyRegExp = new RegExp(/^[a-z0-9][-a-z0-9]*/);
 
 /**
  * Span represents a logical unit of work as part of a broader Trace. Examples
@@ -16,6 +15,25 @@ export default class Span {
     // ---------------------------------------------------------------------- //
     // OpenTracing API methods
     // ---------------------------------------------------------------------- //
+
+    /**
+     * Returns the SpanContext object associated with this Span.
+     *
+     * @return {SpanContext}
+     */
+    context() {
+        if (API_CONFORMANCE_CHECKS) {
+            if (arguments.length !== 0) {
+                throw new Error('Invalid number of arguments');
+            }
+        }
+        let spanContextImp = null;
+        if (this._imp) {
+            spanContextImp = this._imp.context();
+        }
+        // If there's no imp...
+        return new SpanContext(spanContextImp);
+    }
 
     /**
      * Returns the Tracer object used to create this Span.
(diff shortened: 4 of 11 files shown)


src/imp/noop_imp.js
```diff
@@ -1,18 +1,25 @@
...
+NoopSpanContextImp.prototype.getBaggageItem = function (key) {
+}
+
 function NoopSpanImp() {
 }
 NoopSpanImp.prototype.tracer = function () {
     return null;
 }
+NoopSpanImp.prototype.context = function () {
+    // XXX what's the deal here? Why can tracer() return null?
```
bensigelman: note: I left this file in an unfinished state as I wasn't sure I understood how it's meant to fit in.

src/imp/noop_imp.js
```diff
@@ -22,7 +29,7 @@ function NoopTracerImp() {
 }
 NoopTracerImp.prototype.setInterface = function (inf) {
 }
-NoopTracerImp.prototype.startSpan = function (nameOrFields, fields) {
```
bensigelman: I think `nameOrFields` should be omitted here, right? (since this as an impl, and the wrapper calls `startSpan` with exactly one parameter every time)

src/reference.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+/**
+ * Reference pairs a reference type constant (e.g., REFERENCE_CHILD_OF or
+ * REFERENCE_FOLLOWS_FROM) with the SpanContext it points to.
+ *
+ * See the exported childOf() and followsFrom() functions at the package level.
+ */
+export default class Reference {
```
bensigelman: kind of a lot of typing for a `pair<>`...

src/singleton.js
```diff
@@ -72,5 +74,14 @@ export default class Singleton extends Tracer {
 
         // Carrier objects to be exposed at the package level
         this.BinaryCarrier = BinaryCarrier;
+
+        // XXX: what's the best way to export (and document) these two global
```
bensigelman: (just calling this out)

src/span.js
```diff
@@ -18,6 +100,24 @@ export default class Span {
...
+    context() {
+        if (API_CONFORMANCE_CHECKS) {
+            if (arguments.length !== 0) {
+                throw new Error('Invalid number of arguments');
+            }
+        }
+        if (this._imp) {
+            return new Tracer(this._imp.context());
+        }
+        // XXX: what to do here?
```
bensigelman: (this is also known-wrong... I'm more concerned with the user-facing API for this stage of the PR)

src/imp/noop_imp.js
bcronin: `tracer()` can return `null` because this is returning the underlying imp object, not what the caller of the OpenTracing library sees. The caller still sees a OpenTracing `Tracer`. See https://github.com/opentracing/opentracing-javascript/blob/master/src/span.js#L25

Also, for what it's worth, this file is sorely missing a comment that implementation code here is used solely for the unit tests and will never be hit by in normal usage.

src/imp/noop_imp.js
bcronin: Right, the implementation objects have their parameters pre-normalized so that it's easier to create an OpenTracing implementation (and there's less room for error).

src/singleton.js
```diff
@@ -50,7 +51,8 @@ export default class Singleton extends Tracer {
     // Private and non-standard methods
     // ---------------------------------------------------------------------- //
 
-    /* For internal use only:
+    /**
```
bcronin: The comment intentionally did not use the `/**` syntax so that this function would not show up in the generated documentation (the `/**` style doc is what triggers to generate docs or not for the function)

src/span.js
```diff
@@ -6,6 +6,88 @@ let defaultTracer = require('./default_tracer');
...
+ * SpanContext represents Span state that must propagate to descendant Spans
+ * and across process boundaries.
+ *
+ * SpanContext is logically divided into two pieces: the user-level "Baggage"
+ * (see setBaggageItem and getBaggageItem) that propagates across Span
+ * boundaries and any Tracer-implementation-specific fields that are needed to
+ * identify or otherwise contextualize the associated Span instance (e.g., a
+ * <trace_id, span_id, sampled> tuple).
+ */
+export default class SpanContext {
```
bcronin: There can only be one `default` export per file. For consistency, this should likely be in its own file.  I'm a little surprised this didn't generate some kind of error...

src/span.js
```diff
@@ -6,6 +6,88 @@ let defaultTracer = require('./default_tracer');
...
+     *         correspond to a set trace attribute.
+     */
+    getBaggageItem(key) {
+        if (API_CONFORMANCE_CHECKS) {
+            if (arguments.length !== 1) {
+                throw new Error('Invalid number of arguments');
+            }
+        }
+        if (this._imp) {
+            this._imp.getBaggageItem(key);
```
bcronin: This doesn't look correct. This method is supposed to return a string, but doesn't have a return for either the noop or implementation case.

src/singleton.js
bcronin: It's a little odd since, for convenience, the package can be treated either as a singleton or a library (this is somewhat common in the JS world as I understand it).  Therefore, global functions actually should be added as proper methods on this `Singleton` class (i.e. same as `initNewTracer` and `initGlobalTracer`).

src/span.js
bcronin: I'd say this should work similarly to `Tracer.startSpan`. In other words, it should always return a `new SpanContext` and have the Span's `_imp` decide what `spanContextImp` object to pass to the interface object. I.e something like:


```
        let spanImp = null;
        if (this._imp) {
            // < removed some code here for brevity >
            spanImp = this._imp.startSpan(fields);
        }
        return new Span(spanImp);
```
https://github.com/opentracing/opentracing-javascript/blob/master/src/tracer.js#L80

src/span.js
```diff
@@ -106,23 +206,6 @@ export default class Span {
...
-     * Baggage items are copied between all spans, both in-process and across
-     * distributed requests, therefore this feature should be used with care to
-     * ensure undue overhead is not incurred.
-     *
-     * Keys are case insensitive and must match the regular expresssion
-     * `[a-z0-9][-a-z0-9]*`.
-     *
-     * @param {string} key
-     * @param {string} value
-     */
```
bcronin: These comments likely should not be removed as that means these methods will no longer show up in the generated API docs.

src/imp/noop_imp.js
```diff
@@ -1,18 +1,25 @@
+function NoopSpanContextImp() {
```
yurishkuro: nit: why all of these called `Imp`, including the package?  What's wrong with `NoopSpan`, `NoopTracer` ?

src/span.js
```diff
@@ -222,14 +296,11 @@ export default class Span {
...
-     * All Span objects must have finish() called on them before they are
-     * reported to the backend implementation.
-     *
-     * Once `finish()` is called on a Span object, the behavior of all methods
-     * on the object is considered undefined.
+         * Sets the end timestamp and finalizes Span state.
+         *
+         * With the exception of calls to Span.context() (which are always allowed),
+         * finish() must be the last call made to any span instance, and to do
+         * otherwise leads to undefined behavior.
```
bcronin: Tabs?

(By the way, I should add an issue to add `eslint` to this project. It has a `--fix` option that takes care of a lot of things like this automatically)

src/imp/noop_imp.js
yurishkuro: why wouldn't `NoopSpanImp` be initialized with `NoopTracer` and `NoopContext`? I would prefer never to return null from any methods.

src/span.js
```diff
@@ -6,6 +6,88 @@ let defaultTracer = require('./default_tracer');
...
+     * @param {string} key
+     * @param {string} value
+     */
+    setBaggageItem(key, value) {
+        if (API_CONFORMANCE_CHECKS) {
+            if (arguments.length !== 2) {
+                throw new Error('Invalid number of arguments');
+            }
+        }
+        if (this._imp) {
```
yurishkuro: what is a possible use case where `this._imp` is undefined? It feels like these checks just open up the door for subttle & silent bugs.

src/singleton.js
```diff
@@ -72,5 +74,14 @@ export default class Singleton extends Tracer {
...
         // Carrier objects to be exposed at the package level
         this.BinaryCarrier = BinaryCarrier;
+
+        // XXX: what's the best way to export (and document) these two global
+        // helper functions?
+        this.childOf = function(spanContext) {
+            return new Reference(Constants[REFERENCE_CHILD_OF], spanContext);
+        };
+        this.followsFrom = function(spanContext) {
+            return new Reference(Constants[REFERENCE_FOLLOWS_FROM], spanContext);
```
bcronin: I don't think this is right. I assume you mean either `Constants.REFERENCE_CHILD_OF` or `Constants['REFERENCE_CHILD_OF']`.

src/imp/noop_imp.js
bcronin: The NoopSpanImp is misleading per my comment about this being used solely for unit tests.  The consumer of the `opentracing` package will never see `null` returned from any methods.  See the line I linked to in `span.js`.

yurishkuro:
I don't see any unit test files in this change, do we not have any?

src/span.js
bcronin: When `this._imp == null` these classes *are* the no-op implementation. Again, that unit test only file called `noop_imp.js` is unfortunately distracting from that point.

src/span.js
yurishkuro: * I don't think there should be null checks anywhere except maybe in constructors
* the API wrapper objects should be created eagerly, e.g. when `Span(SpanImpl())` is constructed, it immediately captures `Context(impl.context())`, instead of creating a new one every time

src/span.js
yurishkuro: that's a design choice I happen to disagree with. There is a noop impls that can be used instead of nulls.

bensigelman:
@yurishkuro @bcronin I don't think I made it clear enough -- the current rev of this PR is 95% an RFC about what users will see, and 5% questions about how to implement the rest of the change... i.e., don't look carefully at the part of this that callers won't see, at least not until I finish it. (I didn't want to thread everything through until we were ok with the API)

src/imp/noop_imp.js
bensigelman: @yurishkuro my understanding (@bcronin can correct me here) is that the `imp` interface is not quite the same as the user-facing interface in that some params are normalized, etc. (e.g., the startSpan() `fields` param)

bensigelman:
This has been rebased against your changes, @bcronin, and also compiles and passes tests now :)

Please take a look when you get a chance.

bensigelman:
PS: the TravisCI failures (which are not about the code per se AFAICT) are a mystery to me... I will make a trivial change to a comment and re-push to trigger them again since things pass locally.

src/reference.js
```diff
@@ -0,0 +1,38 @@
...
+     *         parent in a REFERENCE_CHILD_OF Reference).
+     */
+    spanContext() {
+        return this._spanContext;
+    }
+
+    /**
+     * Initialize a new Reference instance.
+     *
+     * @param {type} string - the Reference type constant (e.g.,
```
bcronin: `type` and `string` are backwards here

src/span.js
```diff
@@ -18,6 +17,25 @@ export default class Span {
...
+        if (API_CONFORMANCE_CHECKS) {
+            if (arguments.length !== 0) {
+                throw new Error('Invalid number of arguments');
+            }
+        }
+        let spanContextImp = null;
+        if (this._imp) {
+            spanContextImp = this._imp.context();
+        }
+        // If there's no imp...
```
bcronin: nit: this comment isn't accurate as line 35 is run for both branches (though I realize this comment may have been copied & pasted from somewhere else).

src/span.js
```diff
@@ -107,78 +125,6 @@ export default class Span {
...
-    /**
-     * Returns the value for the given baggage item key.
-     *
-     * @param  {string} key
-     *         The key for the given trace attribute.
-     * @return {string}
-     *         String value for the given key, or undefined if the key does not
-     *         correspond to a set trace attribute.
-     */
-    getBaggageItem(key) {
```
bcronin: If documented methods are being removed, let's bump the minor version in `package.json`. 

It might also make sense to have a CHANGELOG or "Versions" section in the README that very briefly documents what the API impact between this version and the prior was - specifically around what the consumer of OpenTracing would need to change in their code to keep everything working.

src/span.js
```diff
@@ -222,14 +168,11 @@ export default class Span {
...
-     * Indicates that the unit of work represented by the span is complete or
-     * has otherwise been terminated.
-     *
-     * All Span objects must have finish() called on them before they are
-     * reported to the backend implementation.
+     * Sets the end timestamp and finalizes Span state.
      *
-     * Once `finish()` is called on a Span object, the behavior of all methods
-     * on the object is considered undefined.
+     * With the exception of calls to Span.context() (which are always allowed),
```
bcronin: This is a bit of an aside: that *any* call is still allowed after `finish()` essentially prevents any sort of object pooling/reuse optimizations that could be done if nothing were valid to call after `finish()`.  I haven't followed / been following the spec discussion closely: is it *necessary* that `context()` is callable after `finish()` (or nice-to-have)?

src/tracer.js
```diff
@@ -18,14 +19,29 @@ export default class Tracer {
...
      *
-     * @param  {string|object} nameOrFields - if the given argument is a
+     * For example:
+     *
+     *     // Start a new (parentless) root Span:
+     *     var parent = Tracer.startSpan('DoWork');
+     *
+     *     // Start a new (child) Span:
+     *     var child = Tracer.startSpan('Subroutine', {
+     *         reference: Tracer.childOf(parent.context()),
```
bcronin: Is using `parent` and a span object directly no longer allowed?

This is a little concerning to me as it adds complexity to the simple case (where I'm new to tracing and am not yet ready for / care about the finer distinctions). In broad terms, I'd argue "mostly everyone" gets what a parent-child relationship is, so `Tracer.startSpan('Subroutine', { parent : parentSpan })` is pretty straightforward.  

By contrast, the snippet above requires two calls to `Tracer`, a sort of redundancy isn both a `childOf` call and a `parent` object, and the next concept of a `context()` (which before I could get away with not knowing about).

I'd argue rather strongly the `parent : <parentObject>` call should still be allowed if that is not a violation of the OpenTracing spec.

test/imp/noop_imp.js
```diff
@@ -1,33 +1,41 @@
...
+
 function NoopSpanImp() {
-}
+    this._context = new NoopSpanContextImp();
+};
 NoopSpanImp.prototype.tracer = function () {
     return null;
 };
+NoopSpanImp.prototype.context = function () {
+    // XXX what's the deal here? Why can tracer() return null?
```
bcronin: Per the comment before, this can return `null` as the SpanContext has an API wrapper. In either case, this comment should be removed.

src/tracer.js
yurishkuro: fwiw, I was upgrading http://github.com/uber-common/opentracing-python-instrumentation and also run into the same annoying problem that we can no longer pass `parent` span. I ended up creating a helper function for now. Even if we allow `childOf()` to accept null reference to parent context (as in https://github.com/opentracing/opentracing-go/pull/96), callers still need to check if `parent` itself is null.

Having said that, allowing `parent=` optional parameter to `startSpan` feels like it's confusing the API. For Go my suggestion is to support `ChildOf(SpanContext)` and `ChildOfSpan(Span)` helpers, both of which allow nil arguments.

bcronin:
The TravisCI tests are failing due to a versioning problem with `eslint` (not the OpenTracing code itself). It's unfortunately a bit of a false negative since it's complaining about a version incompatibility that's not actually incompatible. This is a separate issue and certainly shouldn't hold up this PR.

bcronin:
I created an issue for the TravisCI failures and will look into it now: https://github.com/opentracing/opentracing-javascript/issues/36

src/span.js
bensigelman: (talked about this offline, but TL;DR the `context()` after `finish()` is here for a reason, and it does make it harder to use a pool)

src/tracer.js
bensigelman: @bcronin I can see many (good) arguments for building convenience APIs above this layer... I was also thinking about allowing the caller to pass in a `Span` or a `SpanContext` for references (and using RTTI to convert to a `SpanContext` as needed).

> I'd argue rather strongly the `parent : <parentObject>` call should still be allowed if that is not a violation of the OpenTracing spec.

If we were to do that in the spirit of the current OT nomenclature, perhaps we'd get rid of the `reference: r` shorthand for `references: [r]` and instead have `childOf: spanOrSpanContext`. I.e.,

```
Tracer.startSpan('subroutine', { childOf: parent });
```

Thoughts?

test/imp/noop_imp.js
bensigelman: sorry, this was a bad merge on my part :-/

src/tracer.js
bcronin: That's better :) My goal really is just to make sure the simple case of a very basic parent-child relationship will be simple to someone new to OpenTracing.

bensigelman:
@bcronin @yurishkuro this is ready for another look...

src/singleton.js
```diff
@@ -92,6 +92,7 @@ export default class Singleton extends Tracer {
         for (let key in Constants) {        // eslint-disable-line no-restricted-syntax
             this[key] = Constants[key];
         }
+        this.Reference = Reference;
```
bensigelman: @bcronin this is the proper way to expose this, yes?

bcronin: yup

bensigelman:
@bcronin please take a look...

CHANGELOG.rst
```diff
@@ -0,0 +1,9 @@
+.. :changelog:
+
+History
+-------
+
+0.10.0 (unreleased)
+-------------------
+
+- Move to SpanContext, Reference, and `s/join/extract/g`
```
bcronin: Feel free to include or leave this file out as you want. I think in my original comment I was misusing the standard meaning of "changelog" and actually meant exactly what you have in the README (basically some "porting" tips for API incompatibility changes).  Thanks!

bcronin:
Looks good to me.

CHANGELOG.rst
bensigelman: We have it for `opentracing-python` and I think it's valuable...

PR #34, opened 2016-07-08
Add eslint

bcronin:
## Summary

Adds `eslint` to the project to improve style consistency in the code.

* The rules are a slight one-off from the [airbnb preset](https://www.npmjs.com/package/eslint-config-airbnb) (I don't feel strongly about the particular rules, only that the project be internally consistent in its style)
* Adds a `make lint` rule that'll run `eslint` and auto-fix rule violations that can be auto-fixed (e.g. whitespace, missing semicolons, etc.)
* Runs the rules on the existing code, which resulted in a few small style diffs

Lastly, this also moves the `noop_imp.js` file out of the `src` directory and into `test` since it's a unit test-only file. Having it under `src` was misleading and suggested that code was executed when the interface is not given a specific tracing implementation.


bcronin:
Proceeding with checking this in so there are *some* lint rules being applied. Glad to have the rules updated/tailored by future commits.

PR #41, opened 2016-07-25
Add the HTTP headers carrier format to OT-js

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/98

### DIFF
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -147,25 +147,21 @@ export default class Tracer {
      * within `carrier`. The expected type of `carrier` depends on the value of
      * `format.
      *
-     * OpenTracing defines a common set of `format` values (see FORMAT_TEXT_MAP
-     * and FORMAT_BINARY), and each has an expected carrier type.
+     * OpenTracing defines a common set of `format` values (see
+     * FORMAT_TEXT_MAP, FORMAT_HTTP_HEADERS, and FORMAT_BINARY), and each has
+     * an expected carrier type.
      *
      * Consider this pseudocode example:
      *
      *     var clientSpan = ...;
      *     ...
      *     // Inject clientSpan into a text carrier.
-     *     var textCarrier = {};
-     *     Tracer.inject(clientSpan.context(), Tracer.FORMAT_TEXT_MAP, textCarrier);
+     *     var headersCarrier = {};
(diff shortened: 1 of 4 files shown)


src/tracer.js
```diff
@@ -147,26 +147,22 @@ export default class Tracer {
...
      *     ...
      *     // Inject clientSpan into a text carrier.
-     *     var textCarrier = {};
-     *     Tracer.inject(clientSpan.context(), Tracer.FORMAT_TEXT_MAP, textCarrier);
+     *     var headersCarrier = {};
+     *     Tracer.inject(clientSpan.context(), Tracer.FORMAT_HTTP_HEADERS, headersCarrier);
      *     // Incorporate the textCarrier into the outbound HTTP request header
      *     // map.
-     *     outboundHTTPReq.headers.extend(textCarrier);
+     *     outboundHTTPReq.headers.extend(headersCarrier);
```
bcronin: My comment is a little off-topic as this is both not new to the PR *and* the code comment is explicitly labeled as pseudo-code, but the docs here are a little confusing since `extend` is not a function.  Perhaps the comment could use this:

```js
Object.assign(outboundRequest.headers, headersCarrier);
```

`Object.assign` is at least an ES6 method and supported natively by Node. (Unfortunately, there's no pre-ES6 one-liner for merging associative arrays in JS that I'm aware of that doesn't involve an external lib).

src/tracer.js
```diff
@@ -207,19 +206,17 @@ export default class Tracer {
...
+     * an expected carrier type.
      *
      * Consider this pseudocode example:
      *
      *     // Use the inbound HTTP request's headers as a text map carrier.
-     *     var textCarrier = inboundHTTPReq.headers;
-     *     var wireCtx = Tracer.extract(Tracer.FORMAT_TEXT_MAP, textCarrier);
+     *     var headersCarrier = inboundHTTPReq.headers;
+     *     var wireCtx = Tracer.extract(Tracer.FORMAT_HTTP_HEADERS, headersCarrier);
      *     var serverSpan = Tracer.startSpan('...', Tracer.childOf(wireCtx));
```
bcronin: Also not new, but noticing it: should this be ` Tracer.startSpan('...',  { childOf : wireCtx });`

bcronin:
Looks good.

Two minor comments of the "while you're in the area" variety, that's all :)

PR #42, opened 2016-07-25
Fix inverted operation name length test

bcronin:
The operation name length test was incorrect. It was checking that all operation names are zero length not that they are greater than zero length.


bensigelman:
LGTM

Issue #43, opened 2016-08-16
Add tags to javascript opentracing


oibe:
All of the other opentracing languages have tags defined.  Since javascript is not strictly type checked I don't know if its necessary to have anything more than constants, but its still something we need.

yurishkuro:
In Python the tags are just string constants as well.

oibe:
i'll take a look at that, and make a PR now.

PR #44, opened 2016-08-16
added ext folder with tags

oibe:


oibe:
fixes https://github.com/opentracing/opentracing-javascript/issues/43

yurishkuro:
missing "error" boolean tag, otherwise LGTM.

Please wait for @bensigelman or @bcronin to stamp before merging.

src/ext/tags.js
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2016 Resonance Labs, Inc
```
bensigelman: let's do "The OpenTracing Authors"

bensigelman: actually, we don't have license stuff at the top of the other files. IMO we can just remove this entirely, there's a `LICENSE` in the repo, isn't that sufficient?

src/ext/tags.js
```diff
@@ -0,0 +1,83 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// Here we define standard names for tags that can be added to spans by the
```
bensigelman: let's point to http://opentracing.io/data-semantics/ somewhere

src/ext/tags.js
```diff
@@ -0,0 +1,83 @@
...
+// instrumentation code. The actual tracing systems are not required to
+// retain these as tags in the stored spans if they have other means of
+// representing the same data. For example, the SPAN_KIND='server' can be
+// inferred from a Zipkin span by the presence of ss/sr annotations.
+
+// ---------------------------------------------------------------------------
+// SPAN_KIND hints at relationship between spans, e.g. client/server
+// ---------------------------------------------------------------------------
+export const SPAN_KIND = 'span.kind';
```
bensigelman: elsewhere we define string constants like this:

https://github.com/opentracing/opentracing-javascript/blob/master/src/constants.js#L3

Can we go with that convention just to be consistent?

src/ext/tags.js
oibe: I have no strong opinion on this, but the convention i've seen in many projects is to put this at the top of the file.

yurishkuro:
@bensigelman how do we release JS?

src/ext/tags.js
```diff
@@ -0,0 +1,92 @@
...
+    ERROR : 'error',
+
+    /** ------------------------------------------------------------------------
+    * COMPONENT (string) ia s low-cardinality identifier of the module, library,
+    * or package that is generating a span.
+    * --------------------------------------------------------------------------*/
+    COMPONENT : 'component',
+
+    /** ------------------------------------------------------------------------
+    * SAMPLING_PRIORITY (uint16) determines the priority of sampling this Span.
```
bensigelman: these integer widths don't mean anything in .js... let's just call them all "number"s

bensigelman:
LGTM aside from the above nit.

@yurishkuro https://github.com/opentracing/opentracing-javascript/blob/master/Makefile#L27

bensigelman:
(I can add you on npm, or I can run that for you, either way)

bensigelman:
(Still LGTM aside from the one remaining comment: https://github.com/opentracing/opentracing-javascript/pull/44#discussion_r75021394)

oibe:
I added a brief mention with the opentracing link in tags.js line 26

bensigelman:
thanks!

Issue #45, opened 2016-08-18
Move baggage get/set method from (immutable) SpanContext to Span


yurishkuro:
To bring JS API in parity with other languages, per https://github.com/opentracing/opentracing.github.io/issues/106

Issue #35, opened 2016-07-08
Use normal inheritance instead of interface/implementation coupling


yurishkuro:
@bcronin @bensigelman I've checked with our local JS experts and they suggested a different pattern for defining the API & implementation, which makes more sense to me. The example is:
* interface https://github.com/uber-common/larch/blob/master/base-backend.js
* implementation https://github.com/uber-common/larch/blob/master/logtron-backend.js

The assertions can be removed from prod version via the same pattern already used in this repo, because while they are ok in the above example (where the backend is created only once), they will be expensive for every span/context creation.

bcronin:
That seems fine, though I'm not sure there's such a thing as "normal inheritance" in JavaScript when you're considering not only Node, but also the browser and code that uses `prototype` based inheritance versus down-compiled ES6/7 classes. I'm sure there's some webpack / babel combination that makes it all work.

More of an aside just while I'm, thinking of it, this'll introduce the need for a `getGlobalTracer()` function (or `globalTracer` property, etc.) to ensure generic libraries can get actual object without referencing a specific implementation. Using a true global via `global` would require anyone using `mocha` and consuming an OpenTracing-compatible package to modify the `mocha` default settings.) (Or the notion of a global tracer could be dropped from JS, but that's a difference issue - and one that I think would be a step backwards.)

bcronin:
@yurishkuro / @bensigelman , I finally had some time to remind myself why the `imp` approach was used in the first place.  Sorry for the long-winded notes here, but wanted to capture them somewhere. To be clear, my main goal is to capture the rationale for the `imp` approach - not to convince you that it is The Only Right Way to Do Things or anything crazy like that :)

#### TL;DR

The interface-imp indirection removes initialization order constraints on using the global tracer and also keeps the syntax for using the global tracer short. In short, it reduces the possibility misusing the OpenTracing API.

Below is just details demonstrating this. (And the usual preface: I may be missing something here, so please feel free to point out holes in my logic!)

## The details

#### Concrete example

Let's pretend the OpenTracing Tracer API consists just a `print` method :)

Here's an entry-point file, that's registering a custom OpenTracing implementation:

```javascript
let helpers = require('./helpers.js');
let opentracing = require('./opentracing.js');

opentracing.initGlobal({
    print : function() {
        console.log('hi');
    }
});
let tracer = opentracing.globalInstance();
tracer.print();
helpers.sayHello();
```

And here's the helper file that the author also uses:

```javascript
[...]

PR #46, opened 2016-08-19
Move opentracing-javascript to model where immutable span context is supported.

oibe:
fixes https://github.com/opentracing/opentracing-javascript/issues/45

@bcronin @bensigelman @yurishkuro

### DIFF
--- a/src/span_context.js
+++ b/src/span_context.js
@@ -11,59 +11,6 @@
  * <trace_id, span_id, sampled> tuple).
  */
 export default class SpanContext {
-
-    /**
-     * Sets a key:value pair on this SpanContext that also propagates to future
-     * children of the associated Span.
-     *
-     * setBaggageItem() enables powerful functionality given a full-stack
-     * opentracing integration (e.g., arbitrary application data from a web
-     * client can make it, transparently, all the way into the depths of a
-     * storage system), and with it some powerful costs: use this feature with
-     * care.
-     *
-     * IMPORTANT NOTE #1: setBaggageItem() will only propagate baggage items to
-     * *future* causal descendants of the associated Span.
-     *
-     * IMPORTANT NOTE #2: Use this thoughtfully and with care. Every key and
-     * value is copied into every local *and remote* child of the associated
-     * Span, and that can add up to a lot of network and cpu overhead.
-     *
-     * @param {string} key
-     * @param {string} value
-     */
-    setBaggageItem(key, value) {
-        if (API_CONFORMANCE_CHECKS) {
-            if (arguments.length !== 2) {
-                throw new Error('Invalid number of arguments');
-            }
-        }
-        if (this._imp) {
-            this._imp.setBaggageItem(key, value);
-        }
-    }
-
-    /**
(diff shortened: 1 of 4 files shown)


src/span_context.js
```diff
@@ -11,55 +11,21 @@
...
-     * @param  {string} key
-     *         The key for the given trace attribute.
-     * @return {string}
-     *         String value for the given key, or undefined if the key does not
-     *         correspond to a set trace attribute.
+     * @return {Object} - returns the new SpanContext with key:value pair set
+     * as baggage on the new context.
      */
-    getBaggageItem(key) {
+    withBaggageItem(key, value)  {
```
yurishkuro: withBaggageItem is actually an implementation detail. Does it need to be in the API?  From [this](https://github.com/opentracing/opentracing-javascript/blob/master/src/span.js#L34) it looks like the API method `context()` is always retrieving the current context from the `impl` and wrapping it, so `withBaggageItem` can be handled completely internally to the span implementation.

oibe: Ok.  On another note I see that opentracing-go has forEachBaggageItem on its spanContext.  Do we need that here?

yurishkuro: I would leave it off for another PR. The iterator style may not be idiomatic for JS.

test/imp/noop_imp.js
```diff
@@ -1,8 +1,6 @@
 function NoopSpanContextImp() {
 };
-NoopSpanContextImp.prototype.setBaggageItem = function (key, value) {
-};
-NoopSpanContextImp.prototype.getBaggageItem = function (key) {
+NoopSpanContextImp.prototype.withBaggageItem = function (key, value) {
```
bensigelman: I don't think we need this...

src/span_context.js
```diff
@@ -11,59 +11,6 @@
  * <trace_id, span_id, sampled> tuple).
  */
 export default class SpanContext {
```
bensigelman: I can see an argument for specifically pointing to inject() and extract() in this class comment... since SpanContext no longer has any (!) methods in `ot-javascript`, it's now just a "concept" and a little hard to comprehend without the context of inject and extract.

bensigelman:
/me feels nervous having an interface that has no methods. I would like some guidance from .js natives about the most idiomatic way to represent iteration over baggage items.

I'm fine merging this in the interim, though it would be good to wait for word from @bcronin assuming you can wait until Monday (he is on a short vacay).

oibe:
I don't mind waiting till monday to hear bcronon's opinion.

src/span_context.js
bcronin: For what it's worth, I'm fine with `SpanContext` having no methods. I'd like to do something this week to make the implementation code more inheritance-based than purely implicit per #35 and I think the "empty" class will fit in better then (and if it *doesn't* I can remove it then).

bcronin:
This looks fine to me :)

yurishkuro:
@bcronin @bensigelman are you guys ok to merge this?

bcronin:
Looks good to me

PR #47, opened 2016-08-22
Build Node.js code with Babel, not as a webpack bundle

bcronin:
## Summary

Improves the build process, namely by building the Node.js version of the library using Babel rather than Webpack. The original code built a bundled version of the code using webpack for Node.js mostly due to ignorance on my part about setting up the code correctly (sorry!). I believe what's done in this PR is far more standard.

This also **bumps the version number to `0.12.0`** as there's an API change in here as well: symbol name for the library has been changed from `Tracer` to `opentracing`. (NOTE: I am going to try to make any other API-impacting changes that need to be made *before* pushing this to NPM.)

What this improves:

- The Node.js version is easier to debug as it is a "normal" module rather than a bundle
- Consumers of the library use their own build process to browserify/webpack the files if desired
- The webpack configuration is now a bit simpler since it's not also trying to build Node code

Since there's no longer a Node.js "debug bundle", the following is used instead:

- `require('opentracing')` - the production library, no additional runtime checks
- `require('opentracing/debug')` - the debug library with additional runtime checks

API changes:

- The exported global browser symbol is now `opentracing` rather than `Tracer`
- The `require('opentracing/debug')` is technically an API change (since the debug version is no longer under the `dist` folder)


src/reference.js
```diff
@@ -37,7 +37,7 @@ export default class Reference {
      *        (in which case its .context() is used here).
      */
     constructor(type, referencedContext) {
-        if (API_CONFORMANCE_CHECKS) {
+        if (process.env.NODE_ENV === 'debug') {
```
bcronin: Babel has a plug-in for expanding out environment variables, but not one for arbitrary global constants. Also `NODE_ENV` is a rather standard. It made sense to use this check instead of `API_CONFORMANCE_CHECKS`.

yurishkuro:
@bcronin can you rebase off master to pick up code coverage?

README.md
```diff
@@ -57,13 +57,32 @@ Tracer.initGlobalTracer(TracingBackend.create());
...
+Requiring `opentracing/debug` will include a version of the library with additional runtime checks that are useful for debugging but not desirable for production use.
+
 ## API Documentation
 
 There is a hosted copy of the current generated [ESDoc API Documentation here](https://doc.esdoc.org/github.com/opentracing/opentracing-javascript/).
 
 ## Notes on backwards-incompatible changes
 
-### v0.9.x to v0.10.x
+### v0.11 to v0.12
```
yurishkuro: can we move this to CHANGELOG.md ?

bcronin:
Sure. Do you want to merge #46 first?

lib/ext/tags.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
```
yurishkuro: usually git shows files as moved, is it the case here?

yurishkuro:
I merged #46

lib/ext/tags.js
bcronin: This actually isn't a moved file, it's the file built from `src/ext/tags.js`.  The source file itself, `src/ext/tags.js` is unchanged by this PR.

All the files in `lib` are the down-compiled ES5 JavaScript built from the ES6 source. To the best of my knowledge, it's a standard convention to check-in these built files so consumers of the library do not need to install and run babel.

yurishkuro: I see. I was expecting that ES5 will only be in the packaged version uploaded to npm. But checking it sounds good to, as long as the generator is stable and won't create spurious differences between commits.

lib/ext/tags.js
```diff
@@ -0,0 +1,96 @@
...
+
+// Copyright (c) 2016 Resonance Labs, Inc
```
yurishkuro: @bensigelman mentioned he wanted to change this to OT Authors, still the case?

bcronin: Good catch. Just added a commit to remove it altogether. Nowhere else in the library do we include a copyright notice within the source, so it seemed both simpler and more consistent to remove it.

bcronin:
@yurishkuro , I merged master if you want to give this another look.

yurishkuro:
LGTM. 

I suggest you ignore red coveralls, it's above 1% threshold coverage drop.

lib/ext/tags.js
bensigelman: indeed!

PR #50, opened 2016-08-23
Move to inheritance

bcronin:
**@yurishkuro this is purely an RFC -- obviously not to be merged since it's not complete :)**

I want to make sure this approach is compatible with what you're hoping for before proceeding too far. The gist of it is this:

*Global tracer*

* `require('opentracing/global')` will return a *delegate* wrapper on the global tracer. Users of the global tracer can use this to access the global tracer and avoid package initialization-order concerns. Users who don't care about the global tracer can ignore this package. 
* `require('opentracing')` will return just the global symbols...not the global tracer. (One of those symbols can be `getGlobalTracer` to return the *un-delegated* global tracer if that's desirable.)

*Derived classes*

The base classes provide a "public" method like `startSpan` which does three things (1) checks the API arguments if in debug mode, (2) normalizes the arguments each implementation doesn't have to reimplement the argument normalization, (3) calls a similarly named `_startSpan` method on the class: *this is what the derived class would implement*. 

Of course, if he implementation doesn't care about (1) or (2) they can always just override `startSpan` directly.


bcronin:
Added some work-in-progress `MockTracer` code to (eventually) provide an example tracer implementation + something to use for additional unit testing.

src/span.js
```diff
@@ -189,81 +254,60 @@ export default class Span {
...
 
-    /**
-     * Constructs a new Span object, this method should not be called directly;
-     * Tracer.startSpan() or Tracer.join() should be used instead.
-     */
-    constructor(imp) {
-        this._imp = imp;
+    // By default returns a no-op SpanContext.
+    _context() {
+        return new SpanContext();
```
yurishkuro: I think this should always return the same instance of noop-context, there's no need to keep allocating a new object

src/span.js
```diff
@@ -189,81 +254,60 @@ export default class Span {
...
+    //
+    // The base class could store the tracer that created it, but it does not
+    // in order to ensure the no-op span implementation has zero members,
+    // which allows V8 to aggressively optimize calls to such objects.
+    _tracer() {
+        return noopTracer;
+    }
+
+    // By default does nothing
+    _setOperationName(name) {
```
yurishkuro: is the intention that the implementations (those who want to be wrapped) would implement these private methods? If not, why not stop at the public methods being no-op?

bcronin: Yes, the intention is that derived classes would implement these methods, per line 265.

src/span.js
bcronin: Good point. That said, I haven't been paying attention to the details so much as I primarily wanted to make sure this general refactoring met the requirements you had in mind.

yurishkuro:
@bcronin the overall approach of separating "global" singleton from the API lgtm. But I don't fully understand the purpose of delegating to empty `_methods`. Also, if hypothetically someone wanted to just implement top-level methods directly, couldn't it be done even with the existing delegating API?

@rf could use your expertise here, you mentioned other patterns of doing singletons.

bcronin:
@yurishkuro , the point of delegating to empty `_methods` is (1) and (2) of what I wrote in the summary section of the PR.  Can you clarify what is confusing about what I wrote there?

---
*Derived classes*

The base classes provide a "public" method like `startSpan` which does three things (1) checks the API arguments if in debug mode, (2) normalizes the arguments each implementation doesn't have to reimplement the argument normalization, (3) calls a similarly named `_startSpan` method on the class: this is what the derived class would implement.

---

bcronin:
@yurishkuro , I'll try answering your second question: "Also, if hypothetically someone wanted to just implement top-level methods directly, couldn't it be done even with the existing delegating API?"

No, they could not. The code current in master does not export the `Tracer` class from the package. Everything intentionally went through the delegation so that the library could enforce that arguments were always normalized and it would be impossible to hit an initialization order problem (regardless of what tracer implementation the user was using).  

Just exporting the symbol in the existing code would not be helpful either as the rest of the library always expects the Tracer to be "delegated" and, in short, the rest of the code simply not work.

bcronin:
By the way, the idea of having the "public" `method` call a private `_method` is essentially what's sometimes called the ["Template Method"](https://isocpp.org/wiki/faq/strange-inheritance#private-virtuals) in C++.  The motivation is just as described in the linked to FAQ. The FAQ is about C++, the reasoning is the same (emphasis added):

> When you need to make **specific behavior in a base class customizable** in derived classes, **while protecting the semantics of the interface** (and/or the base algorithm therein), which is defined in public methods that call private virtual methods.

I've spent a non-trivial amount of time working with public APIs with multiple, different implementations in the past and it's certainly my experience that more you can do to protect the semantics of an API from being changed by the underlying implementation, the better.

If the implementations due only implement the `_method` functions then it is a lot less likely that an implementation is going to accidentally or intentionally introduce an extra argument or "special" signature particular to their implementation.  A concrete example of this might be an implementation choosing to return a boolean from `flush()` to indicate some state: if they implement `_flush` this can't happen since `flush()` is enforcing that it does not have a return value. Of course, in JavaScript it's really hard to *enforce* this as the implementor can choose to override the `method` rather than the `_method`, but I'd like to set it up to encourage them not to.

bcronin:
@yurishkuro, I removed the "RFC" from the title as the not implemented parts are now implemented and I'd like to get this merged.  Some updated notes:

* I fixed the unit tests to conform to the API breaking changes
* The low code coverage numbers comes from the fact that the unit tests do not exercise the all the debug argument checking code paths.  In an ideal world, sure they'd be covered, but in practice It's hard to justify adding unit tests to all of these as a good use of time at the moment.
* I added an example/demo via `make example` that uses the `MockTracer` I added to the code
* I updated the README and CHANGELOG to reflect all the changes

bcronin:
Aside: one "regression" I'm not super happy with in moving to an inheritance model is that implementation-specific functions are no longer differentiable in the calling code from API functions. 

```javascript
let tracer = new CustomTracer();

// OpenTracing API call
let span1 = tracer.startSpan('my_span'); 

// Implementation-specific method defined by CustomTracer()
let span2 = tracer.startSpanFromCurrentContext(ctx);
```

This pushes more work to library users (and their code reviewers) to make sure they don't accidentally let implementation-specific functionality sneak into their code bases (which is a real problem in large organizations). Anyway, I think the commentary on this is "so it goes", but wanted at least highlight it before moving on.

bcronin:
FYI, the travis-ci failure is specific to Node v0.12.0. Looking into it. Seems to be something with the Babel generated code.

src/tracer.js
```diff
@@ -1,15 +1,33 @@
...
+ * "public" methods such as `startSpan()` which do two things before calling
+ * the a similarly named worker method (in this case, `_startSpan()`), which the
+ * derived class is expected to implement. The base does two things before
+ * calling the worker method:
+ *
+ * (1) It normalizes the arguments before passing the arguments along to the
+ * derived class. This avoids all N implementations from having to re-implement
+ * error-prone argument manipulation to support different call signatures.  For
+ * example, `inject()` allows either a `Span` or a `SpanContext` to be passed to
+ * the function; `_inject()` will always be passed a `SpanContext`.
```
rf: I'm not sure I understand the benefit here. Why not just have a single versioned API spec to implement to?

src/global_tracer.js
```diff
@@ -0,0 +1,67 @@
...
+// Acts a bridge to the global tracer that can be safely called before the
+// global tracer is initialized. The purpose of the delegation is to avoid the
+// sometimes nearly intractible initialization order problems that can arise in
+// applications with a complex set of dependencies, while also avoiding the
+// case where
+class GlobalTracerDelegate extends Tracer {
+
+    _startSpan(...args) {
+        const tracer = _globalTracer || noopTracer;
+        return tracer._startSpan(...args);
```
yurishkuro: I think this should be calling public methods on the underlying implementation. Then if an application decides to depend on the global singleton, it can work with anything that implements public OT API, not just those implementations that extend it and only implement the `_` methods.

src/tracer.js
```diff
@@ -55,6 +73,7 @@ export default class Tracer {
      * @return {Span} - a new Span object.
      */
     startSpan(nameOrFields, fields) {
```
yurishkuro: since we're breaking the API already, why can't we make it strict, `startSpan(operationName, options)`? It will be consistent with APIs in other languages. The current flexibility seems very superficial, requiring more work from implementations.

src/span.js
```diff
@@ -107,6 +156,7 @@ export default class Span {
      * @return {[type]} [description]
      */
     addTags(keyValuePairs) {
```
yurishkuro: nit: should we clarify that `keyValuePairs` is actually a hash (maybe use a different name?), and not an array of things?

src/span.js
```diff
@@ -180,6 +239,7 @@ export default class Span {
      *         implementation) will be used.
      */
     finish(finishTime) {
+        // Debug-only runtime checks on the arguments
```
yurishkuro: nit: other platforms also accept a collection of logs, but that can be a different PR.

src/tracer.js
```diff
@@ -128,7 +144,11 @@ export default class Tracer {
      * @return a REFERENCE_CHILD_OF reference pointing to `spanContext`
      */
     childOf(spanContext) {
```
yurishkuro: this is, strictly speaking, a static helper function, not a method of the tracer. Is it possible to export those separately, as `opentracing.childOf()`? Then from-scratch tracer implementations (like those written directly in ES5 to avoid babel dependencies) do not have to re-implement it.

src/tracer.js
```diff
@@ -139,7 +159,11 @@ export default class Tracer {
      * @return a REFERENCE_FOLLOWS_FROM reference pointing to `spanContext`
      */
     followsFrom(spanContext) {
```
yurishkuro: same comment as for `childOf`

src/tracer.js
```diff
@@ -263,6 +281,7 @@ export default class Tracer {
      *        was successful.
      */
     flush(done) {
```
yurishkuro: can we remove this from public API?  It was decided against in other languages, and this method does not always make sense, e.g. in case of in-memory only span collectors.

test/unittest.js
```diff
@@ -1,105 +1,185 @@
...
+            expect(new opentracing.Tracer()).to.be.an('object');
         });
 
-        it('should have the required constants', function() {
-            expect(Tracer.FORMAT_TEXT_MAP).to.be.a('string');
-            expect(Tracer.FORMAT_BINARY).to.be.a('string');
-            expect(Tracer.REFERENCE_CHILD_OF).to.be.a('string');
-            expect(Tracer.REFERENCE_FOLLOWS_FROM).to.be.a('string');
+        var tracer = new opentracing.Tracer();
+        var funcs = [
```
yurishkuro: Is it possible to refactor this file to be reusable for testing API compatibility of other implementations? Similar to this: https://github.com/opentracing/opentracing-python/blob/master/opentracing/harness/api_check.py

src/tracer.js
bcronin: Sure. I didn't realize this wasn't in the spec!

bcronin:
Updates:

- Moved `childOf` & `followsFrom` to be standalone functions
- Removed `flush`
- Removed the `nameOrFields`
- Renamed `keyValuePairs` to `keyValueMap`
- `GlobalTracerDelegate` calls the outer methods, not the `_methods`
- Changed the docs to not imply the `_method` functions are necessarily the way to implement a tracer
- Slight refactor & added docs on how to run the API unit tests on a different tracer instance

I'm tempted to remove the `_methods` altogether as well if no one is using them but am lacking a bit on the time/energy to do that right now. I'm still surprised that keeping the user API 1:1 with the implementation API is a goal since that's certainly not the pattern I'm used to seeing on any long-standing, stable API layers, but concede defeat as I seem to have been outvoted on this one :)

bcronin:
@yurishkuro , any thoughts on the updated code or is this good to go?

yurishkuro:
@bcronin :shipit: 

Sorry, been a busy day. Didn't have time for a thorough review, but had a quick look and from your commit comments LGTM. Thanks a lot for `api_compatibility.js`, I'd like to put it to use right away.

Question - how did you get code coverage up? Do you have tests verifying the code that checks invalid parameters?

Also, is `api_compatibility.js` transpiled to es5, if we wanted to test an es5 implementation?

bcronin:
`api_compatibility.js` does **not** run through Babel, so it should be ES5 compliant as-is. I haven't verified this, but that should be the case (and if it's not, it should be relatively easy to make it so).

I believe the code coverage is up is a combination of (1) I added a few additional tests, (2) the changes reduced the number of error checking branches, for example,  removing the `nameOrFields` and `flush`, and (3) the I-don't-know factor as I was a bit surprised that it increased that much :)

PR #53, opened 2016-09-03
Sketch out key-value logging

bensigelman:
Per opentracing/opentracing.github.io#96

This is only a sketch... once we decide on a specific API I will adjust tests and so on.


bensigelman:
(I added a simple usage example)

@bcronin please take a look (before I flesh this out)

src/span.js
```diff
@@ -171,39 +171,54 @@ export default class Span {
...
     /**
-     * Logs a event with an optional payload.
-     *
-     * @param  {string} eventName - string associated with the log record
-     * @param  {object} [payload] - arbitrary payload object associated with the
-     *         log record.
+     * DEPRECATED
      */
     logEvent(eventName, payload) {
         // Debug-only runtime checks on the arguments
```
bcronin: Should if `DEBUG` branch warn about the deprecation?  Something like...

```javascript
 if (process.env.NODE_ENV === 'debug') {
    console.warn('logEvent has been deprecated. Use log() instead.');
     if (arguments.length >= 1 && arguments.length <= 2) {
     ...
```

bcronin:
Makes sense to me :+1:

yurishkuro:
@bensigelman LGTM, can we merge this?

src/span.js
bensigelman: IMO no, but it's just MO :)

bensigelman:
@yurishkuro sure... in general I was waiting on these PRs until the dust had settled in the various languages. I guess that's happened. Merging now...

bensigelman:
Actually, I want to briefly check to make sure the testing stuff that's recently merged is up-to-date.

yurishkuro:
The testing harness is most likely checking the old `log(fields)` API, which did not accept the timestamp as the second arg.

bensigelman:
@yurishkuro @bcronin @oibe I added some stuff in commit 5f43e5c

src/span.js
```diff
@@ -171,39 +171,54 @@ export default class Span {
...
+     *     span.log({
+     *         size: rpc.size(),  // numeric value
+     *         URI: rpc.URI(),  // string value
+     *         payload: rpc.payload(),  // Object value
+     *         "keys can be arbitrary strings": rpc.foo(),
+     *     });
+     *
+     *     span.log({
+     *         "error.description": error.description(),  // numeric value
+     *     }, error.timestampMillis());
```
yurishkuro: Q: what's `error`?

yurishkuro:
LGTM

bensigelman:
Alright, I'm going to merge this. Hopefully not controversial at this point.

src/span.js
bensigelman: This is/was just an imaginary example object... and I'm annoyed that I just merged this since the preceding line has an incorrect "numeric value" comment. I'll fix that in master.

Issue #54, opened 2016-09-06
Opentracing Compatibility tests are inaccurate


oibe:
@bcronin @bensigelman 
1.) In apiCompatibilityChecks the "createTracer" function is never used, so that means I can't import these tests as a client, since my tracer will never be applied.
2.)  The tests assert that errors are thrown when certain conditions are not met, and that certain types are passed in.  My issue with this is that it may be dictating the responsibility of the implementation too much.  For example at Uber a lot of our clients do not throw exceptions, because they would require try/catch which deoptimizes the v8 interpreter, and also there is usually nothing you can do about some of these errors.  (sidenote: There is a larger solution in place so apps, don't crash.)

Essentially I would like to be able to use opentracing compatibility tests without the additional restrictions of typechecking, and error's being thrown.

yurishkuro:
My 2c is that the tests that check for errors to be thrown on invalid arguments do not belong in the API test harness, since those errors are not required by the API. These tests are only testing the specific implementation, which happens to be the default noop tracer. I suggest moving those out to a separate test file.

yurishkuro:
My 2c is that the tests that check for errors to be thrown on invalid arguments do not belong in the API test harness, since those errors are not required by the API. These tests are only testing the specific implementation, which happens to be the default noop tracer. I suggest moving those out to a separate test file.

bcronin:
@oibe, you're right regarding (1). That's a pretty embarrassing oversight :)

Regarding (2), it only throws exceptions in when NODE_ENV="debug" for the very reason that you're describing (more generally, tracing should not interfere with the control flow of the host application, which presumably *should not* be running in NODE_ENV="debug" in production environments). In fact, the code in `lib/` uses babel to drop those checks entirely in the production version (v8 doesn't have to do any optimizations since that code isn't even there). That said, I don't feel strongly about that code  and if you want to remove it, I don't mind.  

All in all, I'm not really using or relying on this API test harness for the LightStep code (which largely explains why I missed that problem that I added `createTracer` be didn't add the necessary calls). As such, since you actually have a specific use case and requirements for the API test harness, do you want to submit a PR to refactor the code to be what you need? That seems like the most direct way of resolving this.

PR #55, opened 2016-09-06
Separated opentracing mock tracer tests from opentracing api

oibe:
compatibility tests.
fixes https://github.com/opentracing/opentracing-javascript/issues/54

### DIFF
--- a/test/api_compatibility.js
+++ b/test/api_compatibility.js
@@ -40,38 +40,38 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
                     expect(tracer[name]).to.be.a('function');
                 });
             });
-        });
 
-        describe('Tracer#startSpan', function() {
-            it('should handle Spans and SpanContexts', function() {
-                var tracer = new opentracing.Tracer();
-                var span = tracer.startSpan('test_operation');
-                expect(function() { tracer.startSpan('child', { childOf : span }); }).to.not.throw(Error);
-            });
-        });
-
-        describe('Tracer#inject', function() {
-            it('should not throw exception on required carrier types', function() {
-                var tracer = new opentracing.Tracer();
-                var spanContext = tracer.startSpan('test_operation').context();
-                var textCarrier = {};
-                var binCarrier = new opentracing.BinaryCarrier();
-                expect(function() { tracer.inject(spanContext, opentracing.FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
-                expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, binCarrier); }).to.not.throw(Error);
-                expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, new Object); }).to.not.throw(Error);
-                expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, {}); }).to.not.throw(Error);
-                expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, { buffer : null }); }).to.not.throw(Error);
-
-                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, binCarrier); }).to.not.throw(Error);
-                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, {}); }).to.not.throw(Error);
-                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : null }); }).to.not.throw(Error);
+            describe('Tracer#startSpan', function() {
+                it('should handle Spans and SpanContexts', function() {
+                    var tracer = new opentracing.Tracer();
+                    var span = tracer.startSpan('test_operation');
+                    expect(function() { tracer.startSpan('child', { childOf : span }); }).to.not.throw(Error);
+                });
             });
 
-            it('should handle Spans and SpanContexts', function() {
-                var tracer = new opentracing.Tracer();
-                var span = tracer.startSpan('test_operation');
-                var textCarrier = {};
-                expect(function() { tracer.inject(span, opentracing.FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
+            describe('Tracer#inject', function() {
+                it('should not throw exception on required carrier types', function() {
+                    var tracer = new opentracing.Tracer();
+                    var spanContext = tracer.startSpan('test_operation').context();
+                    var textCarrier = {};
+                    var binCarrier = new opentracing.BinaryCarrier();
+                    expect(function() { tracer.inject(spanContext, opentracing.FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
+                    expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, binCarrier); }).to.not.throw(Error);
+                    expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, new Object); }).to.not.throw(Error);
+                    expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, {}); }).to.not.throw(Error);
+                    expect(function() { tracer.inject(spanContext, opentracing.FORMAT_BINARY, { buffer : null }); }).to.not.throw(Error);
+
+                    expect(function() { tracer.extract(opentracing.FORMAT_BINARY, binCarrier); }).to.not.throw(Error);
+                    expect(function() { tracer.extract(opentracing.FORMAT_BINARY, {}); }).to.not.throw(Error);
+                    expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : null }); }).to.not.throw(Error);
+                });
+
+                it('should handle Spans and SpanContexts', function() {
+                    var tracer = new opentracing.Tracer();
+                    var span = tracer.startSpan('test_operation');
+                    var textCarrier = {};
+                    expect(function() { tracer.inject(span, opentracing.FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
+                });
             });
         });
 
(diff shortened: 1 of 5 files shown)


bcronin:
@oibe, looks good to me! Thanks!

Note: I added you as a collaborator on the project so you should write access and be able to merge yourself.

bcronin:
@oibe , also let me know your NPM username if you'd like to be added to the "npm-owners" list for the package.  That way `make publish` should allow you to push the new version when needed.

test/api_compatibility.js
```diff
@@ -7,45 +7,17 @@ var expect = require('chai').expect;
...
         describe('Tracer', function() {
             it('should be a class', function() {
-                expect(opentracing.Tracer).to.be.a('function');
-                expect(new opentracing.Tracer()).to.be.an('object');
+                expect(createTracer).to.be.a('function');
+                expect(createTracer()).to.be.an('object');
             });
 
-            var tracer = new opentracing.Tracer();
+            var tracer = new createTracer();
```
yurishkuro: is `new` appropriate here? I would've expected the function itself to return the tracer already.

test/mock_compatibility.js
```diff
@@ -0,0 +1,114 @@
...
+                    expect(opentracing[name]).to.be.a('function');
+                });
+            });
+        });
+
+        describe('Tracer#startSpan', function() {
+            it('should handle Spans and SpanContexts', function() {
+                var tracer = new opentracing.Tracer();
+                var span = tracer.startSpan('test_operation');
+                expect(function() { tracer.startSpan('child', { childOf : span }); }).to.not.throw(Error);
```
yurishkuro: why is this not in the main harness?

test/mock_compatibility.js
```diff
@@ -0,0 +1,114 @@
...
+            var funcs = [
+                'childOf',
+                'followsFrom',
+                'initGlobalTracer',
+                'globalTracer',
+
+            ];
+            _.each(funcs, function(name) {
+                it(name + ' should be a function', function() {
+                    expect(opentracing[name]).to.be.a('function');
```
yurishkuro: should these not be in the main harness? Especially childOf and followsFrom - how are we expecting users to access those functions?

test/mock_compatibility.js
```diff
@@ -0,0 +1,114 @@
...
+                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : null }); }).to.not.throw(Error);
+                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : '' }); }).to.throw(Error);
+                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : 5 }); }).to.throw(Error);
+            });
+            it('should handle Spans and SpanContexts', function() {
+                var tracer = new opentracing.Tracer();
+                var span = tracer.startSpan('test_operation');
+                var textCarrier = {};
+                expect(function() { tracer.inject(span, opentracing.FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
+            });
```
yurishkuro: I think everything that has `to.not.throw` should be in the main harness

test/mock_compatibility.js
```diff
@@ -0,0 +1,114 @@
...
+        });
+
+        describe('Span#finish', function() {
+            it('should return undefined', function() {
+                var tracer = new opentracing.Tracer();
+                var span = tracer.startSpan('test_span');
+                expect(span.finish()).to.be.undefined;
+            });
```
yurishkuro: main harness

test/mock_compatibility.js
```diff
@@ -0,0 +1,114 @@
...
+                        var span = tracer.startSpan('test_span');
+                        span.finish(arg);
+                    }
+                }
+                expect(f(10)).to.not.throw(Error);
+                expect(f(Date.now())).to.not.throw(Error);
+                expect(f('1234567')).to.throw(Error);
+                expect(f([])).to.throw(Error);
+                expect(f({})).to.throw(Error);
+            });
```
yurishkuro: all these - main harness

test/mock_compatibility.js
```diff
@@ -0,0 +1,114 @@
...
+            it('should only accept numbers as an argument', function() {
+                var tracer = new opentracing.Tracer();
+                function f(arg) {
+                    return function() {
+                        var span = tracer.startSpan('test_span');
+                        span.finish(arg);
+                    }
+                }
+                expect(f(10)).to.not.throw(Error);
```
yurishkuro: main harness

test/api_compatibility.js
oibe: It is not appropriate... :( .  It is very inappropriate.

test/api_compatibility.js
```diff
@@ -7,30 +7,14 @@ var expect = require('chai').expect;
...
-            });
-        });
+module.exports = function apiCompatibilityChecks(_createTracer) {
+    var createTracer = _createTracer || function() { return opentracing.Tracer(); };
 
+    describe('OpenTracing API', function() {
         describe('Standalone functions', function() {
             var funcs = [
                 'childOf',
                 'followsFrom',
```
yurishkuro: should these really be on each individual implementation? I would've expected them to be always used from the opentracing namespace.

test/api_compatibility.js
```diff
@@ -56,47 +40,43 @@ module.exports = function apiCompatibilityChecks(createTracer) {
...
-                    expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : '' }); }).to.throw(Error);
-                    expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : 5 }); }).to.throw(Error);
-                });
-                it('should handle Spans and SpanContexts', function() {
-                    var tracer = new opentracing.Tracer();
-                    var span = tracer.startSpan('test_operation');
-                    var textCarrier = {};
-                    expect(function() { tracer.inject(span, opentracing.FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
-                });
+        describe('Tracer#inject', function() {
```
yurishkuro: can we not keep the same nesting/identation to miminize the diff? What do we gain by the change in nesting?

test/api_compatibility.js
```diff
@@ -160,26 +115,13 @@ module.exports = function apiCompatibilityChecks(createTracer) {
...
             var span = tracer.startSpan('test_span');
             var ref = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span.context());
 
             it('should be a class', function() {
                 expect(ref).to.be.an('object');
             });
-            it('should handle Spans and SpanContexts', function() {
-                var tracer = new opentracing.Tracer();
-                var span = tracer.startSpan('test_operation');
-                expect(function() { new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span); }).to.not.throw(Error);
```
yurishkuro: this looks like a good test, why remove it?

test/mock_compatibility.js
```diff
@@ -0,0 +1,87 @@
...
+
+module.exports = function mockCompatibilityChecks() {
+    describe('Mock Opentracing Implementation', function() {
+        describe('Constants', function() {
+            var constStrings = [
+                'FORMAT_TEXT_MAP',
+                'FORMAT_BINARY',
+                'FORMAT_HTTP_HEADERS',
+                'REFERENCE_CHILD_OF',
+                'REFERENCE_FOLLOWS_FROM',
```
yurishkuro: I think this should be just a standalone test. It's not testing "compatibility", it's testing public opentracing API

test/mock_compatibility.js
```diff
@@ -0,0 +1,87 @@
...
+                });
+            });
+        });
+
+        describe('Standalone functions', function() {
+            var funcs = [
+                'childOf',
+                'followsFrom',
+                'initGlobalTracer',
+                'globalTracer',
```
yurishkuro: I think this should be just a standalone test. It's not testing "compatibility", it's testing public opentracing API

test/mock_compatibility.js
```diff
@@ -0,0 +1,87 @@
...
+                expect(function() { tracer.inject(spanContext, opentracing.FORMAT_TEXT_MAP, 5); }).to.throw(Error);
+
+                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : '' }); }).to.throw(Error);
+                expect(function() { tracer.extract(opentracing.FORMAT_BINARY, { buffer : 5 }); }).to.throw(Error);
+            });
+        });
+
+        describe('Span#finish', function() {
+            it('should return undefined', function() {
+                var tracer = new opentracing.Tracer();
```
yurishkuro: even if there's no explicit use case, I think it'd be a nicer design if there was still a factory method to create the tracer that people could theoretically override.

test/mock_compatibility.js
```diff
@@ -0,0 +1,87 @@
+// For the convenience of unit testing, add these to the global namespace
```
yurishkuro: `mock_compatibility` doesn't sound like a right file name. If you move out the global constants & functions tests elsewhere, then the remaining tests here are mostly about parameter validation.

test/api_compatibility.js
```diff
@@ -121,13 +89,8 @@ module.exports = function apiCompatibilityChecks(createTracer) {
             });
```
yurishkuro: we could do it in different PR or in this one, but it seems none of the functions are actually being tested. For example, I would expect tests like:

  * each method should be called at least once by the harness
  * tracer() returns something that has the Tracer methods on it
  * methods like setTag() return something on which setTag can be called again (chaining)
  * get-baggage returns can return what was set with set-baggage (in Python we made this controlled by a parameter to the harness because noop implementation does not store the baggage)

Before we use this harness with our tracer, we need these tests implemented.

test/unittest.js
```diff
@@ -2,8 +2,16 @@
 require('source-map-support').install();
 
 var apiCompatibilityChecks = require('./api_compatibility.js');
+var noopImplementationTests = require('./noop_implementation.js');
+var opentracingAPITests = require('./opentracing_api.js');
+var opentracing = require('../debug.js');
```
yurishkuro: I think it makes sense to run both the debug and non-debug versions of noop tracer through the harness

test/api_compatibility.js
oibe: i'll put it in another diff

Issue #56, opened 2016-09-07
Add assertions based on behavior of opentracing tests


oibe:
From Yuri Shkuro:
> we could do it in different PR or in this one, but it seems none of the functions are actually being tested. For example, I would expect tests like:

> each method should be called at least once by the harness
> tracer() returns something that has the Tracer methods on it
> methods like setTag() return something on which setTag can be called again (chaining)
> get-baggage returns can return what was set with set-baggage (in Python we made this controlled   by a parameter to the harness because noop implementation does not store the baggage)
> Before we use this harness with our tracer, we need these tests implemented.

PR #57, opened 2016-09-08
add behavior based testing to api_compatibility

oibe:


oibe:
So tests currently don't work on the mockTracer because the mockTracer doesn't have methods like "mockTracer.startSpan()" which can be called.  @bcronin can you explain why the mock tracer has keys that represent method names, but not method names that can be called?

@yurishkuro 
Should we add propagation tests?  I didn't really understand the point of the ones in opentracing-python.

oibe:
fixes https://github.com/opentracing/opentracing-javascript/issues/56

yurishkuro:
the point of propagation tests is that you can marshal and unmarshal the span context without loss. But I don't recall that being in the compatibility harness. It should certainly exist for the mock tracer to prove its codecs work.

test/api_compatibility_helpers.js
```diff
@@ -0,0 +1,44 @@
...
+        'context',
+        'tracer',
+        'setOperationName',
+        'finish',
+        'addTags',
+        'log',
+        'logEvent',
+        'setTag'
+    ];
+    return _.every(funcs, function(name) {
```
yurishkuro: I would rather pass the `assert` object to this function and use `isOK` method, so that if a particular function is missing the test output would explicitly list it.

test/api_compatibility_helpers.js
```diff
@@ -0,0 +1,44 @@
...
+
+    return _.every(funcs, function(name) {
+        return tracer[name] instanceof Function;
+    });
+}
+
+exports.hasSpanContextFunctions =
+function hasSpanContextFunctions(spanContext) {
+    var funcs = [
+        'withBaggageItem'
```
yurishkuro: this method is not required by the API, it's implementation detail.

test/api_compatibility.js
```diff
@@ -104,30 +87,66 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
         });
 
-
         describe('Reference', function() {
-            var tracer = createTracer();
-            var span = tracer.startSpan('test_span');
-            var ref = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span.context());
 
             it('should be a class', function() {
+                var ref = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span.context());
```
yurishkuro: this looks like it should be moved to the other test file you have, for opentracing lib itself. It's not checking compatibility of external tracer implementation.

test/api_compatibility.js
```diff
@@ -104,30 +87,66 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+                if (carrierFormatFactoryMap) {
+                    var carrierFactory = carrierFormatFactoryMap[opentracing.FORMAT_TEXT_MAP];
+                    var spanContext = tracer.extract(opentracing.FORMAT_TEXT_MAP, carrierFactory());
+                    assert.isOk(hasSpanContextFunctions(spanContext));
+                }
+            });
+        });
+
+        describe('Span', function() {
+            it('context#should return a span context', function() {
```
yurishkuro: why "context#should" ?

test/api_compatibility.js
```diff
@@ -1,46 +1,48 @@
...
+var hasSpanContextFunctions = require('./api_compatibility_helpers.js').hasSpanContextFunctions;
 
 // Unit testing is done against the debug version of the library as it has
 // additional conformance checks that are optimized out of the production
 // library.  Again, globals are used purely for convenience.
 var opentracing = require('../debug.js');
 
-module.exports = function apiCompatibilityChecks(_createTracer) {
+// TODO(oibe) update documentation to address carrier factory map
+module.exports = function apiCompatibilityChecks(_createTracer, carrierFormatFactoryMap, checkBaggageValues) {
```
yurishkuro: why should `carrierFormatFactoryMap` be provided externally?  The type of carrier is not implementation-dependent.

test/api_compatibility.js
```diff
@@ -104,30 +87,66 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+
+            it('should set baggage and retrieve baggage', function() {
+                span.setBaggageItem('some-key', 'some-value');
+                var val = span.getBaggageItem('some-key');
+                assert.isOk(val);
+                if (checkBaggageValues) {
+                    assert.equal('some-value', val);
+                }
+            });
+        });
```
yurishkuro: what about all other span methods like setTag, log, etc. Even if we cannot validate the return value (sometimes we can, setTag are chainable), we at least can check that those methods are callable without exceptions.

test/api_compatibility.js
oibe: I thought the carrier had to  be a key/value object, and thus it could be some implementation dependent object that some tracing library chose to create.

yurishkuro: What type of object the carrier is depends on the format, nothing else. What that object is populated with is implementation dependent, but you do not need to know what it's populated with if your test start with empty one, do inject, then do extract from the result of inject. For example: https://github.com/uber/jaeger-client-go/blob/master/propagation_test.go#L24

test/api_compatibility.js
```diff
@@ -55,45 +70,23 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
                     expect(span[name]).to.be.a('function');
                 });
             });
 
             describe('Span#finish', function() {
                 it('should not throw exceptions on valid arguments', function() {
-                    var tracer = createTracer();
                     function f(arg) {
                         return function() {
-                            var span = tracer.startSpan('test_span');
```
yurishkuro: don't you want a new span every time? otherwise you're calling finish twice on the same span

test/api_compatibility.js
```diff
@@ -103,30 +96,78 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
             });
         });
 
-        describe('SpanContext', function() {
-            var tracer = createTracer();
-            var span = tracer.startSpan('test_span');
-            var spanContext = span.context();
+        describe('Reference', function() {
+            it('should handle Spans and SpanContexts', function() {
+                expect(function() { new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span); }).to.not.throw(Error);
```
yurishkuro: for consistency with the description, I would also call it with `span.context()`

test/api_compatibility.js
```diff
@@ -103,30 +96,78 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+            });
+        });
+    });
 
-            it('should be a class', function() {
-                expect(spanContext).to.be.an('object');
+    describe('Opentracing implementation compatibility', function() {
```
yurishkuro: why the new scope? how is it different from L34?

test/api_compatibility.js
```diff
@@ -103,30 +96,78 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+        beforeEach(function() {
+            tracer = createTracer();
+            span = tracer.startSpan('test-span');
+        });
+
+        describe('Tracer', function() {
+            _.each(spanFunctions, function(name) {
+                it(name + ' should be a method', function() {
+                    var span = tracer.startSpan('test-span');
+                    expect(span[name]).to.be.a('function');
```
yurishkuro: how is this different from L82?

test/api_compatibility.js
```diff
@@ -103,30 +96,78 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+                });
+            });
+
+            it ('extract returns span context with allowed functions', function() {
+                // pass by default if no factory map was specified.
+                if (carrierFormatFactoryMap) {
+                    var carrierFactory = carrierFormatFactoryMap[opentracing.FORMAT_TEXT_MAP];
+                    tracer.extract(opentracing.FORMAT_TEXT_MAP, carrierFactory());
```
yurishkuro: the description does not match test behavior - there are no assertions in the test

test/api_compatibility.js
```diff
@@ -103,30 +96,78 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+                }
             });
         });
 
+        describe('Span', function() {
+            _.each(spanFunctions, function(name) {
+                it(name + ' should be a method', function() {
+                    expect(span[name]).to.be.a('function');
```
yurishkuro: looks like this is repeated for the 3rd time

test/api_compatibility.js
```diff
@@ -103,30 +96,78 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+                }
             });
 
-            it('should handle Spans and SpanContexts', function() {
-                var tracer = new opentracing.Tracer();
-                var span = tracer.startSpan('test_operation');
-                expect(function() { new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span); }).to.not.throw(Error);
+            it('setTag should be chainable', function() {
+                var newSpan = span.setTag('key', 'value');
+                expect(span['setTag']).to.be.a('function');
```
yurishkuro: I would add a boolean `chainable` to `spanExecutions` table above, and test across the board, since setTag is not the only chainable method (logs and setOperationName should be too)

test/api_compatibility.js
oibe: These tests likely implementations, but the api compatibility tests above assert that functions exist on an object.

test/api_compatibility.js
oibe: Its also asserting that whatever "startSpan" returns has span methods.  Its reusing the code above that validates spans.

test/api_compatibility.js
yurishkuro: it's all API compatibility to me, whether the function exists or returns the right value is still an API conformance, not a test of the implementation (which we can't test anyway)

test/api_compatibility.js
oibe: We should remove this now that OT span context has no functions.

yurishkuro: you can at least test that an object is returned. And you can validate that it's a good spanContext by using it in some other function, like starting a child span with it.

test/api_compatibility.js
yurishkuro: the span in L82 is also created with startSpan, so what's different here?

test/api_compatibility.js
```diff
@@ -1,46 +1,61 @@
...
                 it(name + ' should be a method', function() {
                     expect(tracer[name]).to.be.a('function');
                 });
             });
 
             describe('Tracer#startSpan', function() {
                 it('should handle Spans and SpanContexts', function() {
-                    var tracer = createTracer();
-                    var span = tracer.startSpan('test_operation');
                     expect(function() { tracer.startSpan('child', { childOf : span }); }).to.not.throw(Error);
```
yurishkuro: add another line passing `span.context()`

test/api_compatibility.js
```diff
@@ -103,30 +97,75 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
             });
         });
 
-        describe('SpanContext', function() {
-            var tracer = createTracer();
-            var span = tracer.startSpan('test_span');
-            var spanContext = span.context();
+        describe('Reference', function() {
+            it('should handle Spans and span.context()', function() {
+                expect(function() { new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span); }).to.not.throw(Error);
```
yurishkuro: add a second line that takes `span.context()`

test/api_compatibility.js
```diff
@@ -55,45 +70,24 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
                 });
 
                 it('should handle Spans and SpanContexts', function() {
-                    var tracer = createTracer();
-                    var span = tracer.startSpan('test_operation');
                     var textCarrier = {};
                     expect(function() { tracer.inject(span, opentracing.FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
```
yurishkuro: add another line passing span.context()

test/api_compatibility.js
```diff
@@ -103,30 +97,75 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+                if (checkBaggageValues) {
+                    assert.equal('some-value', val);
+                }
             });
 
-            it('should handle Spans and SpanContexts', function() {
-                var tracer = new opentracing.Tracer();
-                var span = tracer.startSpan('test_operation');
-                expect(function() { new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span); }).to.not.throw(Error);
+            it('setTag should be chainable', function() {
```
yurishkuro: this test is now redundant, it's covered by L151-153

test/api_compatibility.js
```diff
@@ -103,30 +97,75 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
...
+                ], chainable: false},
+                {name: 'logEvent', args: ['eventName', null], chainable: false},
+                {name: 'logEvent', args: ['eventName', {'key': 'value'}], chainable: false},
+                {name: 'finish', args: [], chainable: false},
+                {name: 'setBaggageItem', args: ['key', 'value'], chainable: true},
+                {name: 'getBaggageItem', args: ['key'], chainable: false},
+            ];
+            _.each(spanExecutions, function(a) {
+                it(a.name + 'should not throw exceptions', function() {
+                    var newSpan = span[a.name].apply(span, a.args);
```
yurishkuro: s/newSpan/maybeSpan/

test/api_compatibility.js
```diff
@@ -55,45 +81,59 @@ module.exports = function apiCompatibilityChecks(_createTracer) {
                 });
```
yurishkuro: last 3 checks are validating `extract` method, not `inject` as stated in L76.  Maybe move them to a separate `describe('extract')`?  Also extract is only called with binary format, we could add textmap.

yurishkuro:
One minor suggestion, otherwise LGTM.

bcronin:
@oibe, thanks for this!

PR #62, opened 2017-03-02
Add example for FollowsFrom and improve README

yurishkuro:


src/tracer.js
```diff
@@ -34,6 +34,13 @@ export default class Tracer {
      *         childOf: parent.context(),
      *     });
      *
+     *     // Start a new async (FollowsFrom) Span:
+     *     var child = Tracer.startSpan('Subroutine', {
```
bhs: nit: `s/'Subroutine'/'pipeConsumer'` (or something else that implies `followsFrom`)

Issue #63, opened 2017-03-16
What is the purpose of operationName in startSpan?


robsimmons:
The API documentation suggests that the purpose is to allow a single-argument call to tracer.startSpan

```
tracer.startSpan({
    operationName: "makeRPC",
    tags: {...}
});
```

but the code doesn't actually support this single-argument setup.

yurishkuro:
it's a leftover from previous API. Operation name is a required first argument.

bhs:
@robsimmons thanks for bringing this up. Let me know if my PR clarifies things sufficiently.

robsimmons:
Is `Span.setOperationName` similarly deprecated, and does it similarly act to override the previously given name?

yurishkuro:
it is not deprecated, on the contrary. There are cases when the correct span name is not available at the time of span creation, for example when the span is started by a generic instrumentation in some http framework, where the best that instrumentation can do is to call the span GET or POST, but once the control gets to the real business handler it can override it with something more meaningful like "getCustomerAddress".

robsimmons:
Makes sense! Thanks for clarifying.

PR #64, opened 2017-03-16
Clarify comment for `fields.operationName`

bhs:
Closes https://github.com/opentracing/opentracing-javascript/issues/63

cc @robsimmons @yurishkuro


robsimmons:
Better! It's still unclear what a conforming implementation is expected to do if only one argument (an Object) is provided, or what is supposed to happen if a string first argument and a fields.operationName is provided.

bhs:
@robsimmons PTAL

robsimmons:
👍

yurishkuro:
I'll merge, I think this is not controversial.

PR #68, opened 2017-04-02
Use TypeScript

felixfbecker:
This is a PR to change the implementation from transpiled ES6 code to transpiled TypeScript code.

The biggest advantage is type safety for properties, call signatures, return types and object interfaces, for the code in this repo, implementations and consumers.
Code is transpiled to ES5 code just like before, so for JavaScript consumers nothing changes.
TypeScript users can now use the package with TypeScript and get generated definition files (.d.ts) for free.

As TypeScript is a superset of JavaScript, the actual changes are pretty small. All that's added is a few type declarations and visibilities for `protected` methods that are intended to be overridden by implementations.

TODO:
- [x] Adapt code
- [x] Adapt tests
- [x] Add TSLint
- [x] Remove Babel
- [x] Configure webpack
- [x] Add generated documentation with tsdoc
- [x] Update README


src/tracer.ts
```diff
@@ -1,8 +1,48 @@
...
+
+    /**
+     * a manually specified start time for the created Span object. The time
+     * should be specified in milliseconds as Unix timestamp. Decimal value are
+     * supported to represent time values with sub-millisecond accuracy.
+     */
+    startTime?: number;
+}
+
+export type Format = typeof Constants.FORMAT_BINARY | typeof Constants.FORMAT_TEXT_MAP | typeof Constants.FORMAT_HTTP_HEADERS;
```
yurishkuro: will this restrict the format argument to just these three? It is meant to be open-ended.

felixfbecker: Oh, wasn't aware of that. This would indeed restrict it to these three. An implementation can of course overload the signature.

src/tracer.ts
```diff
@@ -42,28 +82,10 @@ export default class Tracer {
...
-     *        should not modify this object after calling startSpan).
-     * @param {number} [fields.startTime] - a manually specified start time for
-     *        the created Span object. The time should be specified in
-     *        milliseconds as Unix timestamp. Decimal value are supported
-     *        to represent time values with sub-millisecond accuracy.
+     * @param {SpanFields} [fields] - the fields to set on the newly created span.
      * @return {Span} - a new Span object.
      */
-    startSpan(name, fields) {
+    startSpan(name: string, fields: SpanFields = {}): Span {
```
yurishkuro: I'd prefer `s/fields/options/`, it seems to be a common pattern in JS libs and it is more accurate than "fields". Similarly, `s/SpanFields/StartSpanOptions/`

felixfbecker: Totally agree, since only a defined set of properties are accepted.

src/tracer.ts
```diff
@@ -130,7 +151,10 @@ export default class Tracer {
      * @param  {any} carrier - see the documentation for the chosen `format`
      *         for a description of the carrier object.
      */
-    inject(spanContext, format, carrier) {
+    inject(spanContext: SpanContext | Span, format: typeof Constants.FORMAT_TEXT_MAP, carrier: { [key: string]: string }): void;
+    inject(spanContext: SpanContext | Span, format: typeof Constants.FORMAT_BINARY, carrier: { buffer: number[] }): void;
```
yurishkuro: is `number[]` the idiomatic representation of a byte buffer?

src/tracer.ts
```diff
@@ -130,7 +151,10 @@ export default class Tracer {
...
+    inject(spanContext: SpanContext | Span, format: typeof Constants.FORMAT_HTTP_HEADERS, carrier: { [key: string]: string }): void;
+    inject(spanContext: SpanContext | Span, format: Format, carrier: any): void {
```
yurishkuro: Doesn't typescript support generics? Java API uses generics to enforce type parity between format and carrier without having to define overloaded methods. In practice many tracer implementations do a lookup in the internal map for injectors/extractors corresponding to the appropriate format, so forcing them to implement 4 methods is not necessary.

yurishkuro:
I kinda like this (admittedly, I'm biased by my preference for strongly typed over dynamic languages). And the transpiled JS is not horrible at all, it's actually even better than Babel version as it assigns to prototypes directly instead of using a table-driven initialization.

src/tracer.ts
felixfbecker: I wasn't sure because the documentation doesn't specify this. It just says an "Array-like" that carries binary data. I interpreted that as a byte array, eg https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array or NodeJS Buffer (which would be compatible with `number[]` / `Array<number>`)

src/tracer.ts
felixfbecker: TypeScript does support both generic classes and methods. What I wanted to accomplish here was to enforce that if you eg pass `FORMAT_BINARY`, the `carrier` needs to be an object with a `buffer` property.
Overloads in TypeScript are not matched automatically, you just define the overloaded signatures, and then you must define one "generic" implementation that usually does runtime type checks. Implementations do not need to implement the 4 overloads.

felixfbecker:
Glad you like the general direction. I think the source is still easy to understand and contribute to for native JavaScript developers, since it really is just JavaScript + types. Emitting readable JS is one of TS' defined language goals, so you can count on that ;)

tedsuo:
I'm all for moving to TypeScript. Please make sure not to change the API implicitly while lifting up on to it.

felixfbecker:
I addressed the feedback, added typedoc documentation, configured webpack, removed Babel and lib-debug

bhs:
@felixfbecker FYI, I have been time-bankrupt but finally took a look at this.

TL;DR "thank you". I should not have much weight in this discussion since (a) I am just a journeyman-level javascript programmer and (b) the change here is not about OT semantics, but rather js ecosystem best practices. That said, the .ts and .js for the important files is/are all readable.

bhs:
Oh, one request, though: please add something to the README about how both Typescript and non-Typescript javascript programmers should interact with this code (esp the non-Typescript people who might be scared off without some docs pointing them towards `lib/tracer.js` or whatever).

felixfbecker:
I think this is ready for final review. I adapted the tests, updated the readme and removed the compiled JS from git.

I feel like after this is merged it would be a very good opportunity to release v1.

felixfbecker:
Also, in case someone is interested, https://github.com/sourcegraph/javascript-typescript-langserver is completely instrumented using this branch.
It is used together with async/await and Observables and also both injects and extracts contexts into/from JSON RPC messages.

felixfbecker:
Important step needed for the documentation: I've changed the docs generator to typedoc and added a deployment to travis to https://opentracing-javascript.surge.sh. Someone needs to publish this once, then set the secret SURGE_LOGIN and SURGE_TOKEN env vars in the travis settings.

felixfbecker:
ping @yurishkuro @tedsuo @bhs

README.md
```diff
@@ -77,71 +77,65 @@ http.request(opts, function (res) {
...
 
 ### Usage in the browser
 
-The package contains two bundles built with webpack that can be included using a standard `<script>` tag. The library will be exposed under the global `opentracing` symbol:
+The package contains two bundles built with webpack that can be included using a standard `<script>` tag. The library will be exposed under the global `opentracing` namespace:
 
 * `dist/opentracing-browser.min.js` - minified, no runtime checks
 * `dist/opentracing-browser.js` - debug version with runtime checks
 
+### Usage with TypeScript
```
bhs: thank you

felixfbecker:
In theory it's a feature. But I would actually suggest to bump major with this for a couple of reasons:
- I removed runtime type checks
- The lib-debug.js entry point is now missing, so if you imported from that, it won't work anymore
- There are some slight differences in module exports. Babel (incorrectly) exported the whole exports object as the default export, which allows to import with `import opentracing from 'opentracing'` instead of `import * as opentracing from 'opentracing'`. I saw this in the LightStep codebase (the runtime tracer still works though)
- A library is not under semver below 1.0 and considered experimental. I am sure we all would feel more comfortable with a 1.0 release in production, and I feel like this is a good starting point.

Also make sure to configure surge before merging, or the build will fail.

bcronin:
@felixfbecker , just wanted to add that this is awesome work

src/test/api_compatibility.ts
```diff
@@ -0,0 +1,93 @@
+
```
yurishkuro: This is a comment about span_context.ts (which I can't comment on directly).
* `default class` was replaced in other files
* can it be just an interface?

src/test/api_compatibility.ts
```diff
@@ -0,0 +1,93 @@
...
+            tracer = createTracer();
+            span = tracer.startSpan('test-span');
+        });
+
+        describe('Tracer', () => {
+
+            describe('startSpan', () => {
+                it('should handle Spans and SpanContexts', () => {
+                    expect(() => { tracer.startSpan('child', { childOf : span }); }).to.not.throw(Error);
+                    expect(() => { tracer.startSpan('child', { childOf : span.context() }); }).to.not.throw(Error);
```
yurishkuro: add followsFrom?

src/test/api_compatibility.ts
```diff
@@ -0,0 +1,93 @@
...
+            });
+
+            describe('extract', () => {
+                it('should not throw exception on required carrier types', () => {
+                    const textCarrier = {};
+                    const binCarrier = new BinaryCarrier();
+                    expect(() => { tracer.extract(FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
+                    expect(() => { tracer.extract(FORMAT_BINARY, binCarrier); }).to.not.throw(Error);
+                    expect(() => { tracer.extract(FORMAT_BINARY, {}); }).to.not.throw(Error);
+                    expect(() => { tracer.extract(FORMAT_BINARY, { buffer : null }); }).to.not.throw(Error);
```
yurishkuro: not clear why empty and null binary carriers are value combinations

src/tracer.ts
```diff
@@ -0,0 +1,188 @@
...
+     *         ],
+     *     });
+     *
+     * @param {string} name - the name of the operation (REQUIRED).
+     * @param {SpanOptions} [options] - options for the newly created span.
+     * @return {Span} - a new Span object.
+     */
+    startSpan(name: string, options: SpanOptions = {}): Span {
+
+        // Convert fields.childOf to fields.references as needed.
```
yurishkuro: `s/field./options./`

src/test/opentracing_api.ts
```diff
@@ -0,0 +1,79 @@
+
+import { expect } from 'chai';
+import * as opentracing from '../index';
+
+export function opentracingAPITests(): void {
```
yurishkuro: Q: assuming we keep ot.Tracer as a concrete class, why can't these checks be in the api-compatibility file?

src/test/opentracing_api.ts
```diff
@@ -0,0 +1,79 @@
...
+            it('should be a class', () => {
+                const spanContext = span.context();
+                expect(spanContext).to.be.an('object');
+            });
+        });
+
+        describe('Reference', () => {
+            it('should be a class', () => {
+                const ref = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, span.context());
+                expect(ref).to.be.an('object');
```
yurishkuro: can this ever be not an object, given that it's created with `new`?

src/tracer.ts
```diff
@@ -0,0 +1,188 @@
...
+import Span from './span';
+import SpanContext from './span_context';
+
+export interface SpanOptions {
+    /**
+     * DEPRECATED: the name to use for the newly created span. If provided,
+     * overrides the first argument to startSpan(). Provided for
+     * backwards-compatibility.
+     */
+    operationName?: string;
```
yurishkuro: imo we should drop it, especially if we are going for v1.0

README.md
```diff
@@ -42,32 +42,32 @@ Spans:
...
+    res.on('error', err => {
         // assuming no retries, mark the span as failed
         span.setTag(opentracing.Tags.ERROR, true);
-        span.log({'event': 'error', 'error.object': err);
+        span.log({'event': 'error', 'error.object': err, 'message': err.message, 'stack': err.stack});
         span.finish();
     });
-    res.on('data', function (chunk) {
+    res.on('data', chunk => {
         span.log({'event': 'data_received', 'data': chunk});
```
felixfbecker: sure, but the example was there before. I only converted the example to use latest JS language features.

README.md
```diff
@@ -77,71 +77,65 @@ http.request(opts, function (res) {
...
 
 ## OpenTracing tracer implementations
 
 *This section is intended for developers wishing to* ***implement their own tracers***. *Developers who simply wish to* ***use OpenTracing*** *can safely ignore this information.*
 
 ### Custom tracer implementation
 
-Implementations can subclass `opentracing.Trace`, `opentracing.Span`, and the other API classes to build a OpenTracing tracer.
-
-Due to the dynamic nature of JavaScript, implementations can simply implement classes with the same signatures as the OpenTracing classes and use these directly as well (there's no need to subclass).
```
felixfbecker: It depends. If you use TypeScript, and type your method to take an opentracing `Span`, it will work as long as the implementation has the same methods and properties. For JS, it doesn't matter of course. But I didn't get why this needs to be highlighted here, that is simply how JS works and should be obvious to any JS dev. I didn't want to confuse people by explaining the exact circumstances, but rather encourage implementations to subclass the opentracing implementations. What do you mean with going through a proxy class?

scripts/docker_test.sh
```diff
@@ -2,4 +2,4 @@
 
 docker run -t --rm --name test-runner-opentracing-javascript \
```
felixfbecker: I cannot answer that @yurishkuro, I didn't write this. I just adapted it to not use the Makefile

src/binary_carrier.ts
```diff
@@ -5,7 +5,10 @@
  * this class is only one such type of object that can be used.
  */
 export default class BinaryCarrier {
-    constructor(binaryData) {
+
+    public buffer?: ArrayLike<number>;
+
+    constructor(public binaryData?: ArrayLike<number>) {
```
felixfbecker: Agree, but again, I just converted it to TypeScript. It is optional because there are test cases that call this without passing any parameter.

src/reference.ts
```diff
@@ -1,6 +1,9 @@
+import { REFERENCE_CHILD_OF, REFERENCE_FOLLOWS_FROM } from './constants';
 import Span from './span';
 import SpanContext from './span_context';
 
+export type ReferenceType = typeof REFERENCE_CHILD_OF | typeof REFERENCE_FOLLOWS_FROM;
```
felixfbecker: Should I let `type()` return `string` instead?

yurishkuro: I would just use strings, similar to how Format is handled.

In other languages we sometimes were able to define a type that was externally extensible, e.g. in Go. But in Java it's just a string, even though the [Format](https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/propagation/Format.java#L36) is strongly typed.

src/span.ts
```diff
@@ -1,12 +1,14 @@
...
+import Tracer from './tracer';
 
 /**
  * Span represents a logical unit of work as part of a broader Trace. Examples
  * of span might include remote procedure calls or a in-process function calls
  * to sub-components. A Trace has a single, top-level "root" Span that in turn
  * may have zero or more child Spans, which in turn may have children.
  */
-export default class Span {
+export class Span {
```
felixfbecker: I don't really understand your concern. These classes _are_ already the noop implementation. What coupling?
For example, if you look at how I used them here: https://sourcegraph.com/github.com/sourcegraph/javascript-typescript-langserver/-/blob/src/project-manager.ts#L264:1-265:1
I declare each method to take a `Span` that defaults to a noop `Span` instance, and I can pass e.g. a LightStep `Span` just fine because it is compatible.

I totally agree that having all these underscore-prefixed methods is kind of useless, they serve no purpose besides little things like not having to convert `Span | SpanContext` to `SpanContext`. Implementations could just override the public methods. That would be a breaking change though.

What we could do is do a complete restructure, have all these now-classes be interfaces (those will not be translated to JS in any way), and then add classes like `NoopSpan`, `NoopTracer` etc. You would use them like this then:
```ts
import { Span, NoopSpan } from 'opentracing'
function foo(span: Span = new NoopSpan()) {

}

File: src/test/api_compatibility.ts (hunk 6 above)
Comment (felixfbecker) [inline src/test/api_compatibility.ts]: If it's an interface, then it's not an noop implementation anymore.

File: src/test/api_compatibility.ts (hunk 7 above)
Comment (felixfbecker) [inline src/test/api_compatibility.ts]: I only converted this file to TS. Tbh I don't even know what a test is useful for that only tests that a function doesn't throw an exception, it doesn't verify any behaviour at all

File: src/test/api_compatibility.ts (hunk 8 above)
Comment (felixfbecker) [inline src/test/api_compatibility.ts]: Not clear to me either

File: src/span.ts (hunk 18 above)
Comment (yurishkuro) [inline src/span.ts]: By coupling I mean the workarounds with initialization for Noop due to circular dependencies. If `opentracing.Tracer` itself is meant to be a concrete implementation, then it has to be a noop implementation, so there's no reason to have Noop as something separate, it just complicates the library.

I am not too concerned with backwards compatibility given v1.0 label (the external API is still the same). But we can keep some of the shared functionality like normalizing references in StartSpan options. I'd just prefer to see a lot fewer underscore methods, especially when the delegation is completly pass through, as pretty much in all Span methods.

File: src/test/opentracing_api.ts (hunk 10 above)
Comment (felixfbecker) [inline src/test/opentracing_api.ts]: Tbh I don't see any value in these tests. They just verify that exports and methods exist. This is nothing the compiler can't warn you about.

File: src/test/opentracing_api.ts (hunk 11 above)
Comment (felixfbecker) [inline src/test/opentracing_api.ts]: No. This test is pretty useless.

File: src/test/api_compatibility.ts (hunk 7 above)
Comment (yurishkuro) [inline src/test/api_compatibility.ts]: It can be useful for pure JS implementations. It checks that a given implementation satisfies the API, and in some cases checks the behavior, like chaining of methods, and baggage.

File: src/span.ts (hunk 18 above)
Comment (felixfbecker) [inline src/span.ts]: Well, as a TS user, I can only use this as soon as the implementation of my choice adopts it. So I wouldn't say it doesn't matter, but I would rather have a cleaner API.

But the outer API _would_ change. You would have to export the noop implementation under a different name if e.g. `Tracer` is now just an interface (look at my example above).

Noop isn't something separate right now really, all the classes are documented of implementing noop behavior, and I already used them that way

File: src/tracer.ts (hunk 12 above)
Comment (bhs) [inline src/tracer.ts]: +1

**Comment by felixfbecker** [comment]
@yurishkuro I addressed your feedback. While I agree that we could change everything to interfaces, I am not clear about how exactly you would want it to look like.
 - we could have a folder `interfaces` and a folder `noop`, with all noop implementations.
 - what happens to the not-noop implementations, like `Reference`?
 - shall we drop all underscore-prefixed methods then?
 - what happens to normalisation that happened there? Are implementations expected to do those themselves? What do others do here (e.g. opentracing-java)?

This would change the change from a _slight_ breaking change to a major breaking change - which means both more effort I would have to put in and longer adoption time for implementations. The PR is already pretty big by now, so how about we release this as 0.14 and switch to interfaces for 1.0?

File: README.md [hunk 19]
```diff
@@ -77,71 +77,65 @@ http.request(opts, function (res) {
...
 
 See the [OpenTracing website](http://opentracing.io) for general information on contributing to OpenTracing.
 
-The project is built using a `Makefile`. Run:
+The project is written in TypeScript and built using a npm scripts. Run:
 
-* `make build` creates the compiled, distributable code
-* `make test` runs the tests
+* `npm run build` creates the compiled, distributable JavaScript code in `./lib`
+* `npm run watch` incrementally sompiles on file changes
```
yurishkuro: ping

yurishkuro:
I'm ok to merge this and consider other changes separately.

README.md
felixfbecker: Sorry, fixed.

bhs:
@felixfbecker last thing: please actually update the minor version in the package.json and also update the CHANGELOG. After that I'm :thumbsup: to merge this. Thanks again, nearly there...

felixfbecker:
Okay. I would usually do this on the master branch with `npm version minor`, because it commits the version increase in package.json + git tags the version automatically.

Also, as a reminder, for the documentation someone needs to:
- check this branch out
-`npm run typedoc`
-`npm install -g surge`
- cd into the typedoc folder, run `surge`
- publish to the domain `opentracing-javascript.surge.sh`
- get an access token with `surge token`
- add the env vars `SURGE_LOGIN` (your email) and `SURGE_TOKEN` (the generated token) to the secret env vars on Travis

Future updates will be deployed automatically.

bhs:
@felixfbecker ok, sounds like a plan re `npm version minor`. I will try to do that and the remaining steps you suggested when I actually have a block to do it in (i.e., not now). Many thanks again, this PR is a substantial improvement IMO.

bhs:
@felixfbecker you had conflicts against master, but github let me rebase your branch and push to the `felixfbecker/opentracing-javascript` repo (which I have done). I'll try merging and going through the steps you outlined above.

felixfbecker:
Watch out, the changes on master were updates to docblocks and a removal of a dependency - it's very likely that these are reverted by this PR because the diff recreates a lot of files

bhs:
@felixfbecker well, I probably effed that up nicely then – I really only had about 30 mins to do all of this in the midst of a bunch of other stuff. For better or worse I was successful at everything in https://github.com/opentracing/opentracing-javascript/pull/68#issuecomment-296284685 and also the minor version bump. :-/

See http://opentracing-javascript.surge.sh/ and let me know if you need help fixing anything that's broken.

felixfbecker:
master build is broken because the surge deploy is broken:

> Please add SURGE_TOKEN in Travis settings (get your token with 'surge token')
> failed to deploy

I saw you added encrypted values to travis.yml, I don't know how these work, for secret env vars I always just set them in the settings: https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings

I will look at which changes were overridden and reapply them tomorrow, it's late night here already.

bhs:
@felixfbecker 

I fixed the master build – was a stupid yaml formatting mistake.

bhs:
(It's hard to test travis stuff without pushing code – not a great state of affairs)

felixfbecker:
Please next time don't force push the branch, I depended on a specific commit hash from that branch and this broke some builds ❤️

felixfbecker:
0.14.0 is not on npm yet - could you `npm publish` it so I can remove the direct git dependency?

felixfbecker:
I also went through the changes on master and didn't notice any regressions.

bhs:
@felixfbecker 

> Please next time don't force push the branch

Yes, I did a rebase against master for reasons cited above. No way around a force-push post-rebase.

I did run the (npm) tests before pushing anything, FYI – anyway, I will `npm publish` now, I wanted to wait overnight to make sure nothing exploded (since the merge and post-merge process was not smooth sailing).

_time passes..._

https://www.npmjs.com/package/opentracing

felixfbecker:
Well, a `git merge master` would have achieved the same without a force push, but it's all fixed now :)

bhs:
/me has a lot of muscle memory built up around `rebase`, sorry :-/

PR #77, opened 2017-04-22
remove underscore dependency in api test

### LINKED ISSUE #75: `Cannot find module 'underscore'` when using apiCompatibilityChecks
The best way to solve this would be to remove the dependency on underscore, but otherwise the library should depend on underscore.

rochdev:
fixes #75


yurishkuro:
Thanks, @rochdev

PR #80, opened 2017-04-25
Document supported Node version

felixfbecker:
Adds `engine` field to package.json and badge to README


yurishkuro:
The badge comes up as n/a

felixfbecker:
Because the latest version does not have an `engines` field

Issue #78, opened 2017-04-24
Update API compatibility testing documentation


rochdev:
The current documentation says to use this snippet to run API compatibility tests:

```js
const apiCompatibilityChecks = require('opentracing/test/api_compatibility.js');
apiCompatibilityCheck(() => new CustomTracer());
```

However, since version `0.14`, it should be instead:

```js
const apiCompatibilityChecks = require('opentracing/lib/test/api_compatibility.js').default;
apiCompatibilityCheck(() => new CustomTracer());
```

See: https://github.com/opentracing/opentracing-javascript#api-compatibility-testing

rochdev:
@yurishkuro Still missing `.default` since TypeScript doesn't export the default directly.

yurishkuro:
hm, it worked for me, although bombed on another issue #79, which might be related.

Is there a way to make TypeScript export the default?

cc @felixfbecker

rochdev:
Not sure, I know it's also the new behaviour of Babel since version 6. For babel there is a plugin to get the old behaviour back, I don't know about TypeScript. This is related to converting `import` statements to `require`, so a quick and easy solution for node is to simply use `require` directly.

rochdev:
Found a few discussions about this:

https://github.com/Microsoft/TypeScript/issues/2719
https://github.com/Microsoft/TypeScript/issues/3337

rochdev:
I don't see any problems with requiring `.default` as long as it's documented.

felixfbecker:
Yeah, I overlooked this in my PR.

yurishkuro:
Given that README uses `const apiCompatibilityChecks = ...` (ES6 feature), wouldn't the example be better with `import` instead of `require`?
```
import apiCompatibilityChecks from 'opentracing/lib/test/api_compatibility.js';
```

felixfbecker:
I would go after what the NodeJS LTS versions we are targeting support, and they don't support `import` yet (they do support `const`, arrow functions, etc). `.default` has actually been part of the CommonJS spec before NodeJS existed so it's not strictly an ES6 concept :)

bhs:
@felixfbecker @yurishkuro I'm taking a look at open issues... what is the status here? I am not knowledgeable enough about Node to have an opinion, but would like to agree on what we think is best (even if we don't have time to do the work right now).

felixfbecker:
I think this should be fixed in the README on latest master:

> ```ts
[...]

PR #81, opened 2017-04-26
Fix typo in compatibility snippet

DanielMSchmidt:
Hi there,

Thanks for this great project, I am excited to use it. I have found a little typo I wanted to fix real quick 👍 

Thanks for your patience!

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ If `mocha` is being used for unit testing, `test/api_compatibility` can be used 
 
 ```javascript
 const apiCompatibilityChecks = require('opentracing/lib/test/api_compatibility.js');
-apiCompatibilityCheck(() => new CustomTracer());
+apiCompatibilityChecks(() => new CustomTracer());
 ```
 
 ### MockTracer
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,8 +134,8 @@ Implementations can subclass `opentracing.Trace`, `opentracing.Span`, and the ot
 If `mocha` is being used for unit testing, `test/api_compatibility` can be used to test the custom tracer. The file exports a single function that expects as an argument a function that will return a new instance of the tracer.
 
 ```javascript
-const apiCompatibilityChecks = require('opentracing/lib/test/api_compatibility.js');
-apiCompatibilityCheck(() => new CustomTracer());
+const { apiCompatibilityChecks } = require('opentracing/lib/test/api_compatibility.js');
+apiCompatibilityChecks(() => new CustomTracer());
```
yurishkuro: according to https://github.com/opentracing/opentracing-javascript/issues/78, this is also missing `.default` after `require()`, could you add it? Then I can merge.

Thanks!

DanielMSchmidt: You may do both, as it both an named and a default export. Shall I still change it?

yurishkuro: People claim it does not work without `.default` in some later versions (of Babel?). I am not an expert in JS ecosystem. I will merge as is, you may want to reply to that other ticket.

PR #82, opened 2017-05-12
global tracer call delegation uses wrong this

bripkens:
The global tracer delegate uses the delegate as the this context when
calling the configured global tracer. This will break many tracing
implementations. Most notably, this will break all tracing
implementations which subclass the opentracing provided Tracer class.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,12 @@
 History
 -------
 
+Unreleased
+-------------------
+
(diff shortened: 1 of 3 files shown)


bripkens:
Any update on this? This is currently a blocking issue for our opentracing users.

bhs:
Sorry, this just fell through the cracks. Will merge.

bhs:
@bripkens `0.14.1` is up at https://www.npmjs.com/package/opentracing

bripkens:
Thanks @bhs!

PR #85, opened 2017-10-29
Fix how requiring apiCompatibilityChecks

MarckK:
The default export in opentracing/lib/test/api_compatibility.js. is a property on the exports object and thus is accessed with `.default`.
There can be only one default property on the exports object. 
This property does not therefore need to be destructured.


yurishkuro:
thanks!

Issue #89, opened 2017-11-30
Allow Access To Mock Tracer Via Opentracing Import


jhorowitz:
Mock tracer could be incredibly useful for unit tests. I saw a TODO for moving it to its own NPM package but I think it'd be more useful if exposed here. If you think it's a good idea, I'd be happy to make a PR and take ownership of its maintenance. 

What are your thoughts?

yurishkuro:
What do you mean by "Via Opentracing Import"? Isn't it already usable?

jhorowitz:
I mean adding it as an export to https://github.com/opentracing/opentracing-javascript/blob/master/src/index.ts

yurishkuro:
+1

yurishkuro:
fixed by #98

PR #90, opened 2017-11-30
Added the mock_tracer to the opentracing export

jhorowitz:
In reference to https://github.com/opentracing/opentracing-javascript/issues/89

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ History
 Unreleased
 -------------------
 
+- The mock_tracer is now exported by the default opentracing import. It's contained by the `Testing` export.
+
 
 0.14.1
 -------------------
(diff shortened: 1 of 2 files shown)


yurishkuro:
don't know why the build is failing.

is it possible to have a unit test that relies on the new import?

jhorowitz:
I'm not sure what you mean. Do you mean write unit tests for mock_tracer or fool the coverage so the build works? I'm planning on writing unit tests for mock_tracer but I don't have enough bandwidth to do that today.

yurishkuro:
I restarted the build, let's see if it succeeds. `master` branch is green, so no reason it shouldn't work in this PR.

What I meant by the import is if you look at https://github.com/opentracing/opentracing-javascript/issues/91#issuecomment-348313018 the import is done as `import { MockTracer } from '../mock_tracer';`  Instead is should be doing the import via `../index` using the change from this PR.

jhorowitz:
I see. I'll tackle this on Monday

yurishkuro:
Done in #98

Issue #93, opened 2018-01-18
SpanOptions is not exported from the module


eli-jordan:
The SpanOptions type is not exported from the main file. This means implementations cannot use this type.

https://github.com/opentracing/opentracing-javascript/blob/master/src/index.ts#L7

PR #94, opened 2018-01-18
expose span options

eli-jordan:
Fixes https://github.com/opentracing/opentracing-javascript/issues/93

The `SpanOptions` type is not exported from the module. This change simply adds an export to the `index.ts` file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ typedoc/
 dist/
 lib/
 .DS_Store
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ typedoc/
 dist/
 lib/
 .DS_Store
-.idea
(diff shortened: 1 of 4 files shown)


yurishkuro:
Could you also fix the tests? E.g. here the import should from the index, not from `../tracer`, to ensure that these imports are not missed.

https://github.com/opentracing/opentracing-javascript/blob/master/src/test/opentracing_api.ts#L5

eli-jordan:
@yurishkuro Done. I also updated `mock_tracer._startSpan(...)` to use the `SpanOptions` type

.gitignore
```diff
@@ -5,3 +5,4 @@ typedoc/
 dist/
 lib/
 .DS_Store
+.idea
```
yurishkuro: this can go into your `$HOME/.gitignore_global`

eli-jordan: Good tip. I haven't use the global ignore file before. I reverted the change to the `.gitignore`

Issue #96, opened 2018-02-22
export mocktracer in index.js


AloisReitbauer:
Right now Mock Tracer is not exported, although it is part of the npm package. Should we remove it from the npm package or export it. Otherwise people will need to use it with a direct file system reference into node_modules which is weird.

yurishkuro:
it should be exported

AloisReitbauer:
Ok, then we also need to add a dependency to lodash in package.json see issue #95

AloisReitbauer:
I am happy to create a PR if you add me to the project

yurishkuro:
fixed by #98

PR #98, opened 2018-02-24
export mock tracer via standard package

AloisReitbauer:
fixed issue #96

### DIFF
--- a/src/mock_tracer/mock_tracer.ts
+++ b/src/mock_tracer/mock_tracer.ts
@@ -34,15 +34,12 @@ export class MockTracer extends opentracing.Tracer {
         return span;
     }
 
-    protected _inject(span: MockSpan, format: any, carrier: any): void {
-        // throw new Error('NOT YET IMPLEMENTED');
+    protected _inject(span: MockSpan, format: any, carrier: any): never {
+        throw new Error('NOT YET IMPLEMENTED');
     }
 
-    protected _extract(format: any, carrier: any): opentracing.SpanContext {
-        // throw new Error('NOT YET IMPLEMENTED');
-        let context: opentracing.SpanContext;
-        context = new opentracing.SpanContext ();
-        return context;
+    protected _extract(format: any, carrier: any): never {
+        throw new Error('NOT YET IMPLEMENTED');
     }
 
     //------------------------------------------------------------------------//
--- a/src/mock_tracer/mock_tracer.ts
+++ b/src/mock_tracer/mock_tracer.ts
@@ -34,12 +34,15 @@ export class MockTracer extends opentracing.Tracer {
         return span;
     }
 
-    protected _inject(span: MockSpan, format: any, carrier: any): never {
-        throw new Error('NOT YET IMPLEMENTED');
+    protected _inject(span: MockSpan, format: any, carrier: any): void {
+        // throw new Error('NOT YET IMPLEMENTED');
     }
 
-    protected _extract(format: any, carrier: any): never {
-        throw new Error('NOT YET IMPLEMENTED');
+    protected _extract(format: any, carrier: any): opentracing.SpanContext {
+        // throw new Error('NOT YET IMPLEMENTED');
+        let context: opentracing.SpanContext;
+        context = new opentracing.SpanContext ();
+        return context;
     }
 
     //------------------------------------------------------------------------//
--- a/src/test/api_compatibility.ts
+++ b/src/test/api_compatibility.ts
@@ -5,6 +5,7 @@ import { BinaryCarrier, FORMAT_BINARY, FORMAT_TEXT_MAP, Reference, REFERENCE_CHI
 export interface ApiCompatibilityChecksOptions {
     /** a boolean that controls whether or not to verify baggage values */
     checkBaggageValues?: boolean;
+    checkInjectBehaviour?: boolean;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


AloisReitbauer:
Hmm, should we add a test for MockTracer to increase coverage?

src/test/mocktracer_implemenation.ts
```diff
@@ -0,0 +1,58 @@
+
+import { expect } from 'chai';
+import {initGlobalTracer, MockTracer } from '../index';
+import { MockSpan } from '../mock_tracer';
+
+export function mockTracerimplementationTests(createTracer = () => new MockTracer()): void {
```
yurishkuro: why define `createTracer` as argument?

src/test/unittest.ts
```diff
@@ -2,9 +2,13 @@
 require('source-map-support').install();
 
 import apiCompatibilityChecks from './api_compatibility';
+import mocktracerImplementationTests from './mocktracer_implemenation';
 import noopImplementationTests from './noop_implementation';
 import opentracingAPITests from './opentracing_api';
 
+// basic tests for the mock tracer
+mocktracerImplementationTests ();
```
yurishkuro: running apiCompatibilityChecks on MockTracer would be more useful

src/test/mocktracer_implemenation.ts
```diff
@@ -0,0 +1,58 @@
...
+            describe('Memory usage', () => {
+                it('should not report leaks after setting the global tracer', () => {
+                    initGlobalTracer(createTracer());
+                });
+            });
+        });
+    });
+}
+
+export default mockTracerimplementationTests;
```
yurishkuro: looks like it's exported twice?

src/test/unittest.ts
AloisReitbauer: Will port them over. Makes sense

src/test/mocktracer_implemenation.ts
AloisReitbauer: This seems to be a copy and paste issue. It could be directly in the test case as well.

src/test/mocktracer_implemenation.ts
AloisReitbauer: did it the same way as nooptracer does

AloisReitbauer:
Ok, API tests would work out of the box. However, some tests are failing due to the current MockTracer implementation

```
  3 failing
  1) OpenTracing API Compatibility Tracer inject should not throw exception on required carrier types:
     AssertionError: expected [Function] to not throw 'Error' but 'Error: NOT YET IMPLEMENTED' was thrown
      at Context.<anonymous> (src/test/api_compatibility.ts:42:109)
  2) OpenTracing API Compatibility Tracer inject should handle Spans and SpanContexts:
     AssertionError: expected [Function] to not throw 'Error' but 'Error: NOT YET IMPLEMENTED' was thrown
      at Context.<anonymous> (src/test/api_compatibility.ts:49:102)
  3) OpenTracing API Compatibility Tracer extract should not throw exception on required carrier types:
     AssertionError: expected [Function] to not throw 'Error' but 'Error: NOT YET IMPLEMENTED' was thrown
      at Context.<anonymous> (src/test/api_compatibility.ts:58:97)
```
The reason is. that the simply throw an not implemented error. 

...
    protected _inject(span: MockSpan, format: any, carrier: any): never {
        throw new Error('NOT YET IMPLEMENTED');
    }

    protected _extract(format: any, carrier: any): never {
        throw new Error('NOT YET IMPLEMENTED');
    }
...

So. this can be simply commented out or should there be a specific implementation. The other option is to add them to the optional API tests.

src/mock_tracer/mock_tracer.ts
```diff
@@ -34,12 +34,15 @@ export class MockTracer extends opentracing.Tracer {
...
+        // throw new Error('NOT YET IMPLEMENTED');
     }
 
-    protected _extract(format: any, carrier: any): never {
-        throw new Error('NOT YET IMPLEMENTED');
+    protected _extract(format: any, carrier: any): opentracing.SpanContext {
+        // throw new Error('NOT YET IMPLEMENTED');
+        let context: opentracing.SpanContext;
+        context = new opentracing.SpanContext ();
+        return context;
```
yurishkuro: I don't think this is a good change - the implementation is incorrect anyway, so throwing exception is more honest. Otherwise it will just confuse people trying to use this mock tracer.

yurishkuro:
Mock tracer that does not support inject/extract is pretty useless as a tool for testing instrumentation. It should at least support HTTP_HEADERS format. But that's a question for another PR. I think your first commit was in better state to be merged.

AloisReitbauer:
I agree, still this was is the current implementation. I will create a new issue for that one work on getting this one closed first. 

Reverted the code throw the exception again. 
Added an option in the test to ignore inject behaviour for now
Added a small API check for report method, to ensure code coverage.

src/test/api_compatibility.ts
```diff
@@ -5,6 +5,7 @@ import { BinaryCarrier, FORMAT_BINARY, FORMAT_TEXT_MAP, Reference, REFERENCE_CHI
 export interface ApiCompatibilityChecksOptions {
     /** a boolean that controls whether or not to verify baggage values */
     checkBaggageValues?: boolean;
+    checkInjectBehaviour?: boolean;
```
yurishkuro: by default these checks should be on, so a better name for the option is `skipInjectExtractTests`

src/test/api_compatibility.ts
```diff
@@ -36,29 +37,35 @@ export function apiCompatibilityChecks(createTracer = () => new Tracer(), option
...
-                    expect(() => { tracer.inject(spanContext, FORMAT_BINARY, binCarrier); }).to.not.throw(Error);
-                    expect(() => { tracer.inject(spanContext, FORMAT_BINARY, {}); }).to.not.throw(Error);
+                    if (options.checkInjectBehaviour) {
+                        const spanContext = span.context();
+                        const textCarrier = {};
+                        const binCarrier = new BinaryCarrier([1, 2, 3]);
+                        expect(() => { tracer.inject(spanContext, FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
+                        expect(() => { tracer.inject(spanContext, FORMAT_BINARY, binCarrier); }).to.not.throw(Error);
+                        expect(() => { tracer.inject(spanContext, FORMAT_BINARY, {}); }).to.not.throw(Error);
+                    }
```
yurishkuro: does Node have a way to mark tests as skipped? It should not succeed silently.

src/test/mocktracer_implemenation.ts
```diff
@@ -0,0 +1,26 @@
+
+import { expect } from 'chai';
+import {MockTracer } from '../index';
+
+export function mockTracerimplementationTests(): void {
```
yurishkuro: can we remove this `export`?

src/test/api_compatibility.ts
```diff
@@ -14,7 +15,7 @@ export interface ApiCompatibilityChecksOptions {
  * @param {object} createTracer - a factory function that allocates a tracer.
  * @param {object} [options] - the options to be set on api compatibility
  */
-export function apiCompatibilityChecks(createTracer = () => new Tracer(), options: ApiCompatibilityChecksOptions = {}): void {
+function apiCompatibilityChecks(createTracer = () => new Tracer(), options: ApiCompatibilityChecksOptions = {checkInjectBehaviour: true}): void {
```
yurishkuro: `{checkInjectBehaviour: true}` - this is not sufficient. If someone overrides the Options for another reason, e.g. to skip baggage checks, the default for `checkInjectBehaviour` again reverts to false. That's why it should be a property with a negative name, i.e. `skipXyzChecks`, so that only when someone explicitly opts-in the checks will be skipped.

Same applies to baggage, we should rename it to `skipBaggageChecks`

src/test/api_compatibility.ts
```diff
@@ -35,7 +36,7 @@ export function apiCompatibilityChecks(createTracer = () => new Tracer(), option
             });
 
             describe('inject', () => {
-                it('should not throw exception on required carrier types', () => {
+                options.checkInjectBehaviour ? it : it.skip('should not throw exception on required carrier types', () => {
```
yurishkuro: did you mean 

`(options.checkInjectBehaviour ? it : it.skip)('should not throw exception on required carrier type...`?

in the current format it looks like it just returns the `it` function without invoking it.

src/test/api_compatibility.ts
```diff
@@ -44,21 +45,21 @@ export function apiCompatibilityChecks(createTracer = () => new Tracer(), option
                     expect(() => { tracer.inject(spanContext, FORMAT_BINARY, {}); }).to.not.throw(Error);
                 });
 
-                it('should handle Spans and SpanContexts', () => {
+                options.checkInjectBehaviour ? it : it.skip('should handle Spans and SpanContexts',  () => {
                     const textCarrier = {};
                     expect(() => { tracer.inject(span, FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
-                    expect(() => { tracer.inject(span.context(), FORMAT_TEXT_MAP, textCarrier); }).to.not.throw(Error);
+                    expect(() => { tracer.inject(span.context(), FORMAT_TEXT_MAP, textCarrier); });
```
yurishkuro: if you're removing `.to.not.throw(Error);`, then what is the expectation that's being set?

src/test/api_compatibility.ts
AloisReitbauer: I tried to keep as much as it was originally. I had the same idea, but thought it keep things as they are as much as possible. It seems, however, that this class needs more rework anyways.

src/test/api_compatibility.ts
AloisReitbauer: That was not by intention.

src/test/api_compatibility.ts
AloisReitbauer: Yes.

src/test/api_compatibility.ts
yurishkuro: I'd say if the old way is broken, no reason to continue on the same path.

src/test/api_compatibility.ts
```diff
@@ -3,8 +3,9 @@ import { assert, expect } from 'chai';
 import { BinaryCarrier, FORMAT_BINARY, FORMAT_TEXT_MAP, Reference, REFERENCE_CHILD_OF, Span, Tracer } from '../index';
 
 export interface ApiCompatibilityChecksOptions {
-    /** a boolean that controls whether or not to verify baggage values */
-    checkBaggageValues?: boolean;
+    /** a boolean that controls whether or not to verify certain API functionality */
+    skipCheckBaggageValues?: boolean;
+    skipCheckInjectBehaviour?: boolean;
```
yurishkuro: I am being a bit pedantic about the naming, because it constitutes a public API.
  * skipCheckBaggageValues used to control only reading the baggage, now it's a larger scope.
  * skipCheckInjectBehaviour controls both inject & extracy
  * I would prefer to have a uniform naming scheme for these flags, e.g. `skip{Something}Checks`, resulting in
    * skipBaggageChecks
    * skipInjectExtractChecks

PR #105, opened 2018-04-04
edit OT-JS license info so that GitHub recognizes it

eirinikos:
(All CNCF projects should use the Apache License 2.0 - see opentracing/specification#61 for context.)

- Please correct me if I'm wrong, but it seems like the original intent was to license OT-JavaScript as Apache 2.0, anyway (as mentioned in the [README](https://github.com/opentracing/opentracing-javascript#license)).
- However, the [LICENSE file](https://github.com/opentracing/opentracing-javascript/blob/master/LICENSE) itself contains the text for the MIT License, and a copyright reference to `2016 Resonance Labs, Inc`.
- In any case, GitHub currently detects the OT-JavaScript license as MIT, not Apache 2.0.
- Could someone clarify the reason for referencing the Apache license in `README` and using the MIT license in the `LICENSE` file?

If this commit really does represent a switch from MIT to Apache, then we'd like to give previous contributors a 7-day period to air any objections.

(If you have any concerns or feedback, please post a comment here by April 11. Thanks!)

CCing @dankohn @bhs 

CCing previous contributors: @bcronin @oibe @yurishkuro @bripkens @skzinc @felixfbecker @rochdev @eli-jordan @aroc @AloisReitbauer @jhorowitz @strongliang

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,19 +1,201 @@
-Copyright (c) 2016 Resonance Labs, Inc
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


felixfbecker:
Note that package.json also declares MIT as license.

austinlparker:
This has been sitting for a while with the appropriate approvals and no feedback to the negative, so I'll go ahead and merge it.

PR #107, opened 2018-06-23
update dependencies and ci node version

rochdev:
This PR updates all dependencies to the latest versions. It also bumps the version used by Travis to the latest LTS Node version. This was done because a lot of the tools have dropped support for Node 4.x at this point. `mocha` has been kept to the old version since newer versions don't support Node 0.x.


.travis.yml
```diff
@@ -1,14 +1,16 @@
 language: node_js
 
 node_js:
-- 4
+- 8
```
yurishkuro: I have always used Node 6 to run Babel, would it still work?

I think we should add something to the README, I don't see it mentioning which Node version must be used for building the module.

rochdev: I went directly to Node 8 with Travis for future-proofing, but Node 6 will definitely work. I will add a note to the documentation.

austinlparker:
I think this has been open long enough for people to give feedback, seeing none, I'm going to go ahead and merge. Thanks @rochdev!

Issue #112, opened 2018-08-14
ScopeManager for in-process context propagation


sebnow:
The JavaScript implementation should follow suite of languages like Java and Python and implement the proposed ScopeManager specification. This would make managing the active span significantly easier.

RFC: https://github.com/opentracing/specification/blob/master/rfc/scope_manager.md
Python: https://github.com/opentracing/opentracing-python/pull/63
Java: https://github.com/opentracing/opentracing-java/pull/188

yurishkuro:
It's a bit tough given that there are no thread locals in Javascript. There might be something coming up from https://github.com/nodejs/diagnostics

pauldraper:
Node.js diagnostics are limited to the Node.js runtime.

AWS X-Ray uses https://github.com/othiym23/node-continuation-local-storage which in turn is built on https://github.com/othiym23/async-listener. It has built-in support for Node.js APIs.

I would recommend using https://github.com/angular/zone.js which is robust with a long history. It also has an associated TC39 proposal https://github.com/domenic/zones (in early staged). It has built-in support for both web browser and Node.js APIs.

yurishkuro:
Good to have domain experts 🙂

Are there minimum Node version limitations in those cls libraries?

junminstorage:
I used node-contination-local-storage in the tutorial a while ago:
https://github.com/junminstorage/opentracing-tutorial/blob/master/nodejs/lesson04/solution/hello-context.js

junminstorage:
node-cls support was developed in 2012, supported 0.12.x, and is very lower level library, The latest nodejs has async hook
for any application level library we want to use like request, or promise lib e.g. bluebird, you need to patch it with a cls version e.g. cls-bluebird which is what I did in the tutorial. Back then I was trying to mimic the python contextManager in nodeJS, but didn't like sample code I ended up with, I forgot the reasons why though.

Puneeth-n:
Also https://www.npmjs.com/package/cls-hooked

rochdev:
Here is a summary of the situation, along with the version support information asked above.

## Node modules

The following Node modules exist and are popular enough to be mentioned here.

### [async-listener](https://github.com/othiym23/async-listener)

#### Overview

This is the oldest one and probably the most used one for Node applications. It is used by the very popular [continuation-local-storage](https://github.com/othiym23/node-continuation-local-storage).

#### Node support

[...]

PR #113, opened 2018-08-19
ScopeManager

pauldraper:
Addresses #112 

See https://github.com/opentracing/specification/blob/f7ca62c9/rfc/scope_manager.md

## Continuation passing style

I first tried adhering to the draft RFC (see work [here](https://github.com/rivethealth/opentracing-javascript/commit/13fc56cd9fd5ad45d02a1f2a8454570f0db381d1)), but it got clumsy, particularly for the Zone.js implementation. Just about every JS continuation library uses continuation passing style (Zone.js, Node.js domains, continuation-local-storage, etc.), and I believe it's the correct choice. I added my findings to https://github.com/opentracing/specification/issues/126.

## Support

Unfortunately, there just isn't a great universal, out-of-the-box CLS solution.

There are two things that can be addressed by CLS: native async syntax (async/await), and common APIs (e.g. fs.readFile).

The are a surprising number of issues with current solutions. https://github.com/othiym23/node-continuation-local-storage is widely used, but can't be used with async/await which in 2018 is increasingly common. https://github.com/gms1/node-async-context can be, but (so far) doesn't have a lot of adoption and requires Node 8+. https://github.com/angular/zone.js/ has related TC39 proposal, but it's been stalled for two years.

In the end, the JS community has not figured out CLS (yet). I added a few implementations with very minimal dependencies that between them have fairly broad coverage. (See README.md.)

Implementations:
* AsyncHookSpanManager - uses Node.js async_hooks
* AsyncWrapSpanManager - uses async-hook-jl which uses Node.js AsyncWrap
* ZoneSpanManager - uses Zone.js

In the future, I expect the situation to improve. But this seems a good start, these wrappers are rather simple (~50 lines each).


pauldraper:
Update: looks like https://github.com/RisingStack/jaeger-node and https://github.com/RisingStack/opentracing-auto use async_hooks.

src/tracer.ts
```diff
@@ -88,6 +144,11 @@ export class Tracer {
                 options.references = [childOf];
             }
             delete(options.childOf);
+        } else if (!options.references) {
```
rmfitzpatrick: Java and Python have an "ignore active span" option to prevent this automatic reference.  Any chance you'd want to follow suite?

README.md
```diff
@@ -105,6 +105,64 @@ const tracer = opentracing.globalTracer();
...
+    }, 0);
+    (async () => {
+        tracer.activeSpan() // span
+    })();
+});
+tracer.activeSpan() // null
+```
+
+JavaScript support for continuation local storage is a patchwork.
+This library includes simple wrappers around a few common APIs for continuation contexts.
```
rochdev: This should be selected automatically for the user according to the environment. This will avoid the user needing extensive knowledge of all the possible ways to do context propagation in JavaScript.

I've mentioned how I think this should be selected at the bottom of https://github.com/opentracing/opentracing-javascript/issues/112#issuecomment-433982906

README.md
```diff
@@ -105,6 +105,64 @@ const tracer = opentracing.globalTracer();
...
+tracer.activate(span, () => {
+    tracer.activeSpan() // span
+    setTimeout(() => {
+        tracer.activeSpan() // span
+    }, 0);
+    (async () => {
+        tracer.activeSpan() // span
+    })();
+});
+tracer.activeSpan() // null
```
rochdev: Should these be on `tracer.scopeManager()` instead?

README.md
```diff
@@ -105,6 +105,64 @@ const tracer = opentracing.globalTracer();
...
+
+This uses the Node.js [async_hooks](https://nodejs.org/api/async_hooks.html) API, available in Node.js 8.1.x+.
+
+```js
+import { AsyncHookSpanManager } from 'opentracing/async_hook_span_manager';
+
+new AsyncHookSpanManager();
+```
+
+**AsyncWrapSpanManager**
```
rochdev: Since `AsyncWrap` cannot have multiple hooks, and has many bugs, possible crashes and memory leaks in older Node versions, I would probably skip it entirely and add something based on a fixed fork of `async-listener` as described in https://github.com/opentracing/opentracing-javascript/issues/112#issuecomment-433982906

rochdev: FWIW, in my experience it's also slower than `async-listener`.

src/tracer.ts
```diff
@@ -52,6 +55,60 @@ export class Tracer {
...
+    /**
+     * Starts a span, activates the span, executes the async function, and finishes the span.
+     *
+     * @param {string} name the name of the operation.
+     * @param {SpanOptions} options options for the newly created span.
+     * @param {Function} f the function to execute during span activation.
+     * @return {A} the return value of the executed function.
+     * @template A
+     */
+    async runSpanAsync<A>(name: string, options: SpanOptions, f: () => Promise<A>): Promise<A> {
```
rochdev: Not sure I understand why this is necessary as `await` can be used on any promise which can be returned by `runSpan` as well.

src/tracer.ts
```diff
@@ -52,6 +55,60 @@ export class Tracer {
...
+    /**
+     * Starts a span, activates the span, executes the function, and finishes the span.
+     *
+     * @param {string} name the name of the operation.
+     * @param {SpanOptions} options options for the newly created span.
+     * @param {Function} f the function to execute during span activation.
+     * @return {A} the return value of the executed function.
+     * @template A
+     */
+    runSpan<A>(name: string, options: SpanOptions, f: () => A): A {
```
rochdev: This name sounds a bit weird to me, as this method doesn't run a span.

src/tracer.ts
pauldraper: Yes, this should

src/tracer.ts
pauldraper: Other suggestion?

It starts and stops (runs?) the span.
It runs the function, with a new active span for its duration.

src/tracer.ts
pauldraper: It's necessary if you want to stop the span once the promise is resolved.

rochdev: Why not simply detect if the returned value is a promise in `runSpan` and handle it there?

src/tracer.ts
rochdev: I think I would actually remove it completely for the first version, to keep the API surface as small as possible. Then more helpers can be added over time if necessary and discussed individually.

src/tracer.ts
```diff
@@ -52,6 +55,60 @@ export class Tracer {
...
+            span.finish();
+        }
+    }
+
+    /**
+     * Gets the span manager.
+     *
+     * @return {SpanManager} - the span manager, which may be a noop.
+     */
+    spanManager(): SpanManager {
```
rochdev: I think it this could be called simply `scope()`. It makes sense to call `tracer.scope().span()` and `tracer.scope().run()` because these actions are indeed on the current scope.

src/tracer.ts
```diff
@@ -52,6 +55,60 @@ export class Tracer {
     // OpenTracing API methods
     // ---------------------------------------------------------------------- //
 
+    /**
+     * Gets the currently active span.
+     *
+     * @return the active Span. This is a shorthand for tracer.spanManager().active().
+     */
+    activeSpan(): Span | null {
```
rochdev: Similar to my comment below, I don't see the point of this since the alternative is not any more involved.

src/async_wrap_span_manager.ts
```diff
@@ -0,0 +1,56 @@
...
+asyncHookJl.enable();
+
+let currentId: number | null = 1;
+asyncHookJl.addHooks({
+    pre: (id: number) => currentId = id,
+    post: () => currentId = null // in case of bugs
+});
+
+/**
+ * ScopeManger using Node.js {@link async_hooks|https://nodejs.org/api/async_hooks.html}
```
pierlucg-xs, line 14:
```suggestion
 * ScopeManager using Node.js {@link async_hooks|https://nodejs.org/api/async_hooks.html}
```

PR #118, opened 2018-12-11
Add support for knowing if a global tracer has been registered

MikeGoldsmith:
Adds support for knowing if a global tracer has been registered via a global variable including tests.

NOTE: Python currently does not prevent you re-registering a global tracer multiple times like C# and Java does.


MikeGoldsmith:
Any input on this commit? @austinlparker @yurishkuro @carlosalberto

PR #119, opened 2018-12-19
Build improvements

felixfbecker:
- Use files in package.json instead of .npmignore
- Don't publish tsconfig.json, prevents error in VS Code
- Publish src so debugging with source maps work
- Update TypeScript and emit declarationMaps
- Add webpack-cli to dependencies and fix webpack config (didn't build)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
   "engines": {
     "node": ">=0.10"
   },
+  "files": [
+    "src",
+    "lib",
+    "dist"
+  ],
   "license": "Apache-2.0",
   "licenses": [
     {
@@ -12,7 +17,7 @@
     }
   ],
   "main": "lib/index.js",
-  "typings": "lib/index.d.ts",
+  "types": "lib/index.d.ts",
   "scripts": {
     "prepare": "npm run build",
     "lint": "tslint -t msbuild -c tslint.json 'src/**/*.ts'",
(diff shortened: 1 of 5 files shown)


felixfbecker:
@yurishkuro could you give this a review?

package.json
```diff
@@ -4,6 +4,11 @@
   "engines": {
     "node": ">=0.10"
   },
+  "files": [
+    "src",
```
yurishkuro: why do we need src in the artifact?

package.json
```diff
@@ -4,6 +4,11 @@
...
+    "lib",
```
yurishkuro: what is lib?

felixfbecker: It's the output dir of TypeScript, where compiled JavaScript goes.

https://github.com/opentracing/opentracing-javascript/blob/062505c9327c964c2925caaed9aa994ce5684f5e/tsconfig.json#L11

package.json
felixfbecker: As mentioned in the PR description, without the source step debugging doesn't work as the source the sourcemaps point to cannot be found. The alternative would be inlining sources into the source maps (base64).

styfle: Another alternative is to avoid publishing both source and source maps.

package.json
yurishkuro: how is that different from dist?

felixfbecker: `src` has `.ts` files. `dist` has `.js`, `.js.map`, `.d.ts` and `.d.ts.map` files (result of compiling `.ts` fields in `src`).
Note that that the only thing new here is generation of the `.d.ts.map` files and inclusion of `src` in the npm package.

You can compare https://unpkg.com/opentracing@0.14.3/dist/ to https://github.com/opentracing/opentracing-javascript/tree/master/src if you're curious about the differences of the files.

yurishkuro:
would be good to have someone more versed in js / ts to have a look.

austinlparker:
lgtm

felixfbecker:
Since this was approved by two people, could this be merged?

I would really like to get rid of this logspam:

image

PR #122, opened 2019-02-20
Add config files to .npmignore

styfle:
These files do no need to be published to npm


felixfbecker:
A better solution that wouldn't erode with the next added config file would be to delete `.npmignore` and specify the `files` field in package.json

felixfbecker:
I did that in https://github.com/opentracing/opentracing-javascript/pull/119

PR #124, opened 2019-04-03
modify example to use ts-node

jordo1138:
found to run 
`npm run example`
that the package.json path was wrong as was the reference to demo.js instead of demo.ts
got it working locally only by using ts-node


jordo1138:
my bad forgot the run build would transpile everything,

PR #127, opened 2019-05-29
Fixed #126 - MockTracer._allocSpan is private and can't be overwritten

### LINKED ISSUE #126: MockTracer._allocSpan is private and can't be overwritten
MockTracer._allocSpan is private and can't be overwritten - so trace can't be fully customized. Should be protected.

oranoran:
changed _allocSpan() to protected


yurishkuro:
Resolves #126

PR #154, opened 2020-11-04
Fix circular dependency in mock_context

jdmoody:
There seems to be a legitimate circular dependency here that causes problems when I build this library using Snowpack:
1. ./index.ts imports ./mock_tracer/index.ts
2. ./mock_tracer/index.ts imports ./mock_tracer/mock_tracer.ts
3. ./mock_tracer/mock_tracer.ts imports ./mock_tracer/mock_context.ts
4. ./mock_tracer/mock_context.ts imports ./index.ts

I _think_ it should be fixed by simply importing `SpanContext` directly rather than importing the full ./index.ts.

### DIFF
--- a/src/mock_tracer/mock_context.ts
+++ b/src/mock_tracer/mock_context.ts
@@ -1,11 +1,11 @@
-import * as opentracing from '../index';
+import {SpanContext} from '../span_context';
 import MockSpan from './mock_span';
 
 /**
  * OpenTracing Context implementation designed for use in
  * unit tests.
  */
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):
Thanks!

jdmoody:
Thanks for the quick review + merge! 🙏  

I'd love to get a patch version with this out, if possible. I believe anyone trying to use opentracing with Snowpack will run into this, otherwise.

yurishkuro:
Released as `v0.14.15`

Issue #161, opened 2021-03-27
Type checks won't always work in Reference creating functions 


treble-snake:
The interface allows passing both Span and SpanContext as an argument for creating childOf or followsFrom references, both in [helper functions](https://github.com/opentracing/opentracing-javascript/blob/master/src/functions.ts) and the [Reference constructor](https://github.com/opentracing/opentracing-javascript/blob/master/src/reference.ts#L40).

Example:
```
export function childOf(spanContext: SpanContext | Span): Reference {
    // Allow the user to pass a Span instead of a SpanContext
    if (spanContext instanceof Span) {
        spanContext = spanContext.context();
    }
    return new Reference(Constants.REFERENCE_CHILD_OF, spanContext);
}
```

But unless I'm missing something, the `instanceof` check won't always work in runtime. 

For example, I'm working with one of the popular implementations - [jaeger-client-node](https://github.com/jaegertracing/jaeger-client-node). It's [not even directly inherits](https://github.com/jaegertracing/jaeger-client-node/blob/master/src/span.js#L20) `opentracing.Span` class, so the check doesn't work. Then the span object instead of the context object gets into reference, and I end up with an error that took some time to figure out.

Do you think it makes sense to perform a more runtime-proof check, let's say:
```
export function childOf(spanContext: SpanContext | Span): Reference {
    // Allow the user to pass a Span instead of a SpanContext
    if ('context' in spanContext) {
        spanContext = spanContext.context();
    }
    return new Reference(Constants.REFERENCE_CHILD_OF, spanContext);
}
```

I can prepare a PR if this looks reasonable.

yurishkuro:
I wonder why jaeger types don't extend...

Your proposal sounds ok, but I would add the extra check instead of replacing.

treble-snake:
Yeah, no idea, they don't extend opentracing classes, just implement the interfaces.

An extra check could look like this I suppose:
```
if (spanContext instanceof Span || 'context' in spanContext) {
    spanContext = spanContext.context();
}
```

But feels a bit redundant actually.

yurishkuro:
Doesn't seem redundant, there could be a wrapping implementation of SpanContext that has a `context` member.

I would do it the way you showed. Compliant implementations should probably extend Span and short-circuit the evaluation.

PR #162, opened 2021-03-27
Handle non-extending Span classes in Reference constructor

treble-snake:
The interface allows passing both Span and SpanContext as an argument
for creating References.
But the `instanceof` check won't always work in runtime, for example, for cases
when Span implementation does not extend the opentracing.Span class, but just
implements the interface.
The popular `jaeger-client-node` library does exactly that.

This produces hard-to-track errors.

I added additional checks to distinguish between Span and SpanContext instances.

Resolves: #161


src/functions.ts
```diff
@@ -26,9 +22,5 @@ export function childOf(spanContext: SpanContext | Span): Reference {
  * @return a REFERENCE_FOLLOWS_FROM reference pointing to `spanContext`
  */
 export function followsFrom(spanContext: SpanContext | Span): Reference {
-    // Allow the user to pass a Span instead of a SpanContext
-    if (spanContext instanceof Span) {
-        spanContext = spanContext.context();
-    }
```
treble-snake, line 32: Looks like these checks are made in the Reference constructor anyway, thus are duplicated here.

src/test/fixtures/ExtendingSpan.ts
```diff
@@ -0,0 +1,16 @@
+import Span from '../../span';
+import SpanContext from '../../span_context';
+
+/**
+ * Span implementation for unit tests. Extends opentracing.Span.
+ */
+export class ExtendingSpan extends Span {
```
yurishkuro, line 7: could simply use MockSpan

src/test/fixtures/NonExtendingSpan.ts
```diff
@@ -0,0 +1,85 @@
+import * as noop from '../../noop';
+import SpanContext from '../../span_context';
+import Tracer from '../../tracer';
+
+/**
+ * Span implementation for unit tests. Does not extend opentracing.Span, but
+ * implements the interface
+ */
+export class NonExtendingSpan {
```
yurishkuro, line 9: this seems a bit of an overkill to test one check, it could've been done with just an anonymous dict. Why maintain all this boilerplate that doesn't add anything useful?

treble-snake: Thanks for the helpful suggestions, I got rid of the fixtures indeed :)

src/reference.ts
```diff
@@ -1,6 +1,23 @@
 import Span from './span';
 import SpanContext from './span_context';
 
+const getContext = (contextOrSpan: SpanContext | Span): SpanContext => {
+    if (contextOrSpan instanceof SpanContext) {
+      return contextOrSpan;
+    }
```
yurishkuro, line 7: This doesn't seem necessary.

src/reference.ts
```diff
@@ -1,6 +1,23 @@
 import Span from './span';
 import SpanContext from './span_context';
 
+const getContext = (contextOrSpan: SpanContext | Span): SpanContext => {
```
yurishkuro, line 4:
```suggestion
const toContext = (contextOrSpan: SpanContext | Span): SpanContext => {
```

src/reference.ts
treble-snake: This covers a rare, but possible case when a class, extending SpanContext, has its own "context" method for some reason. Do you think it's not worth covering?

yurishkuro (approved):

yurishkuro:
Thanks

treble-snake:
It's been a pleasure :)

yurishkuro:
@treble-snake it's possible this change introduced a bug, see https://github.com/opentracing/opentracing-javascript/issues/172

